PR #12, opened 2021-04-28
Add support for TagOwners and Tests in ACL configuration

### LINKED ISSUE #11: TagOwners -> tagowners = error
Hi,

Access Controls TagOwners looks like this:
```json
  "TagOwners": {
    "tag:tailscale-relay": [
      "group:admins"
    ]
  },
```

But I get such an error for ACL:
```
│ Error: Failed to set ACL
│ 
│   with tailscale_acl.current_acl,
│   on main.tf line 19, in resource "tailscale_acl" "current_acl":
│   19: resource "tailscale_acl" "current_acl" {
│ 
| ports: tag not found: "tag:tailscale-relay" (400)
```

I think the problem is that all the json keys are downcased by terraform tailscale provider like `tagowners`
so tailscale can not find a key `TagOwners` with defined tags 
and fails with an error.

davidsbond:
Fixes #11

Adds support for both the `Tests` and `TagOwners` fields in ACL configuration. These were previously not included
in the request to the ACL endpoint. Updates the acceptance test to include a full ACL and adds a test to ensure
JSON marshalling of the ACL works as expected.

Signed-off-by: David Bond <davidsbond93@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,9 +29,9 @@ install: build
 	mkdir -p ~/.terraform.d/plugins/${HOSTNAME}/${NAMESPACE}/${NAME}/${VERSION}/${OS_ARCH}
 	mv ${BINARY} ~/.terraform.d/plugins/${HOSTNAME}/${NAMESPACE}/${NAME}/${VERSION}/${OS_ARCH}
 
-test: 
-	go test -i $(TEST) || exit 1                                                   
-	echo $(TEST) | xargs -t -n4 go test $(TESTARGS) -timeout=30s -parallel=4                    
+test:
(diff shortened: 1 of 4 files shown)


davidsbond:
Had to raise a question on the forum with regards to capitalisation in the field names, I'm assuming it's `tagowners`, but it could be camel cased `tagOwners`

https://forum.tailscale.com/t/acl-behaviour-when-using-json-instead-of-hujson/601

davidsbond:
Based on a quick test, it seems to be entirely lowercase

PR #27, opened 2021-08-16
docs: added the provider configuration env variables

mvisonneau:
Simply add a couple of existing env variables to configure the provider in the documentation.


davidsbond (approved):
Hey @mvisonneau thanks for the contribution! Looks like a good change to me

PR #34, opened 2021-08-24
Implement devices datasource

### LINKED ISSUE #32: New datasource tailscale_devices
I would be able to have a Machines list

Adding a new datasource named `tailscale_devices` with some argument like : 
`name_prefix` or `regex`
And having  a list of Machine IDs as ids argument

**Additional context**

> data "tailscale_devices" "sample_devices" { 
>         regex = "^server-.*"
> }

> ["71427992728242151", "72824215171427992"]

davidsbond:
This commit implements the `tailscale_devices` data source and allows for
a `name_prefix` argument to filter the list of devices.

Closes #32

### DIFF
--- a/docs/data-sources/devices.md
+++ b/docs/data-sources/devices.md
@@ -1,0 +1,31 @@
+---
+page_title: "devices Data Source - terraform-provider-tailscale"
+subcategory: ""
+description: |-
+"The devices data source describes a list of devices in a tailnet.
+---
+
+# Data Source `devices`
+
+The devices data source describes a list of devices in a tailnet.
+
+## Example Usage
+
+```terraform
+data "tailscale_devices" "sample_devices" {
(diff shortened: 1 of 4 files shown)


davidsbond:
@feraudet was this what you had in mind? Is the `name_prefix` enough or do you need a regular expression also?

feraudet:
> @feraudet was this what you had in mind? Is the `name_prefix` enough or do you need a regular expression also?

name_prefix is sufficient for my use case, but probably regex would be useful for someone else

davidsbond:
> > @feraudet was this what you had in mind? Is the `name_prefix` enough or do you need a regular expression also?
> 
> name_prefix is sufficient for my use case, but probably regex would be useful for someone else

In that case, I'll just keep `name_prefix` for now and can implement a regex one if someone needs it. Thanks!

PR #51, opened 2021-11-08
Set user as a computed attribute, not a required attribute

### LINKED ISSUE #50: User field set as a required argument, rather than computed
**Describe the bug**
@johnae raised an issue in #49 that the `user` field in the `device` and `devices` data sources have been set to `required` rather than `computed`. This means that users of v0.5.0 will need to set the `user` field as a blank string for those resources to work correctly.

**To Reproduce**

Try to use the `device` or `devices` data source without setting the `user` field to blank, see an error.

**Expected behaviour**

No error should occur using those data sources based on the value of the `user` field.

 - Provider Version v0.5.0

**Additional context**
@michael-careplanner has offered to raise a PR to fix this

michael-careplanner:
Fixes bug introduced in https://github.com/davidsbond/terraform-provider-tailscale/pull/49

Fixes #50

### DIFF
--- a/tailscale/data_source_device.go
+++ b/tailscale/data_source_device.go
@@ -20,8 +20,8 @@ func dataSourceDevice() *schema.Resource {
 			},
 			"user": {
 				Type:        schema.TypeString,
-				Required:    true,
 				Description: "The user associated with the device",
+				Computed:    true,
 			},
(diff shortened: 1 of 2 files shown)


davidsbond (approved):
LGTM, TYVM

PR #58, opened 2021-12-22
switch resource type for DeviceSubnetRoutes.routes to Set

### LINKED ISSUE #57: DeviceSubnetRoutes resource detecting spurious changes
**Describe the bug**
With my list of subnet routes on a device, the resource provider for DeviceSubnetRoutes detects changes when none have occurred, due to differences in ordering of the list as returned by the API.

**To Reproduce**
1. Create config with a DeviceSubnetRoutes resource with a number of routes and apply it
2. Change order of the routes in the config and re-apply; (potentially just re-applying without changing order may also reproduce. in my testing, order of routes returned by the API does not appear to be sorted or necessarily fully deterministic)
3. Get a diff with changes in the routes on the server

**Expected behaviour**
Changes in order of the routes either in the terraform definition or as returned by the API should not result in the resource showing as changed.

 - OS: Linux 
 - Terraform Version: 1.0.4
 - Provider Version: 0.5.1

**Additional context**
Simultaneous PR with fix: #58

victorluft:
This PR changes DeviceSubnetRoutes to have TypeSet instead of TypeList.

TypeList requires the items to be in a specific order; TypeSet does not. Using TypeList can result in unnecessary change detection if output returned by TS API (which does not appear to return in a fully deterministic or sorted order) is in a different order than input. It also makes diffs more difficult to read if e.g. you add a route mid-list and it detects all routes after that as changed.

In my testing, at least, this change appears to migrate cleanly from the previous version with TypeList.

Fixes #57

### DIFF
--- a/tailscale/resource_device_subnet_routes.go
+++ b/tailscale/resource_device_subnet_routes.go
@@ -22,7 +22,7 @@ func resourceDeviceSubnetRoutes() *schema.Resource {
 				Description: "The device to set subnet routes for",
 			},
 			"routes": {
-				Type: schema.TypeList,
+				Type: schema.TypeSet,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
@@ -52,7 +52,7 @@ func resourceDeviceSubnetRoutesRead(ctx context.Context, d *schema.ResourceData,
 func resourceDeviceSubnetRoutesCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 	client := m.(*tailscale.Client)
 	deviceID := d.Get("device_id").(string)
-	routes := d.Get("routes").([]interface{})
+	routes := d.Get("routes").(*schema.Set).List()
 
 	subnetRoutes := make([]string, len(routes))
 	for i, route := range routes {
@@ -70,7 +70,7 @@ func resourceDeviceSubnetRoutesCreate(ctx context.Context, d *schema.ResourceDat
 func resourceDeviceSubnetRoutesUpdate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 	client := m.(*tailscale.Client)
 	deviceID := d.Get("device_id").(string)
-	routes := d.Get("routes").([]interface{})
+	routes := d.Get("routes").(*schema.Set).List()
 
 	subnetRoutes := make([]string, len(routes))
 	for i, route := range routes {


davidsbond:
Hey @victorluft, thanks for raising this fix for the issue you've opened. I wasn't aware of the ordering issue but it's great you have a simple solution. 

Looks like some tests are failing as the value returned from [Get](https://github.com/davidsbond/terraform-provider-tailscale/pull/58/files#diff-8248a0e44d32b274a01068ad5dae1501f8a3611e5958baaf94c65143671fffb2R55) is no longer an `[]interface{}`, but a `*schema.Set` now.

If you've got the time please modify the code so that creating/updating works without a panic. You can use `make test` within the repository root to run tests

victorluft:
> Hey @victorluft, thanks for raising this fix for the issue you've opened. I wasn't aware of the ordering issue but it's great you have a simple solution.
> 
> Looks like some tests are failing as the value returned from [Get](https://github.com/davidsbond/terraform-provider-tailscale/pull/58/files#diff-8248a0e44d32b274a01068ad5dae1501f8a3611e5958baaf94c65143671fffb2R55) is no longer an `[]interface{}`, but a `*schema.Set` now.
> 
> If you've got the time please modify the code so that creating/updating works without a panic. You can use `make test` within the repository root to run tests

Ahh yup, my bad; only tested to make sure reading worked. Will have an update up shortly.

davidsbond:
@victorluft thanks. My suggestion would be to either switch to `TypeSet` and use the `Set.List()` function to get the `[]interface{}`, or create a diff suppression function for the subnet routes like we do with the [acl](https://github.com/davidsbond/terraform-provider-tailscale/blob/master/tailscale/resource_acl.go#L26) resource

victorluft:
> @victorluft thanks. My suggestion would be to either switch to `TypeSet` and use the `Set.List()` function to get the `[]interface{}`, or create a diff suppression function for the subnet routes like we do with the [acl](https://github.com/davidsbond/terraform-provider-tailscale/blob/master/tailscale/resource_acl.go#L26) resource

done.

davidsbond:
Perfect, this LGTM, many thanks

davidsbond:
@victorluft I've just released 0.6.0, please give it a try and hopefully your issue is resolved. The release also includes a new resource for managing your tailscale authentication keys, if you're keen it would be handy if you could try that out too.

Thanks again and happy holidays!

victorluft:
yup, with 0.6.0 it works for me!

quick test of the tailnet auth keys seems to be working as well; looking forward to integrating that into some of my infra definitions.

PR #60, opened 2021-12-26
Add tags support to auth keys

michaelbeaumont:

It lets users preselect ACL tags to use for the machines authenticated by a key.

See https://tailscale.com/kb/1068/acl-tags/#generate-an-auth-key-with-an-acl-tag

### DIFF
--- a/tailscale/resource_tailnet_key.go
+++ b/tailscale/resource_tailnet_key.go
@@ -27,6 +27,15 @@ func resourceTailnetKey() *schema.Resource {
 				Description: "Indicates if the key is ephemeral.",
 				ForceNew:    true,
 			},
+			"tags": {
+				Type: schema.TypeSet,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Optional:    true,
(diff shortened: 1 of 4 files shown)


tailscale/resource_tailnet_key.go
```diff
@@ -27,6 +27,15 @@ func resourceTailnetKey() *schema.Resource {
 				Description: "Indicates if the key is ephemeral.",
 				ForceNew:    true,
 			},
+			"tags": {
+				Type: schema.TypeList,
```
davidsbond, line 31: I think this should probably be `TypeSet` instead of `TypeList`, so that duplicate tags aren't allowed.

davidsbond (commented):
Thanks for raising this, it's handy functionality. Could you please update the documentation for the new field on the resource, as well as including tags in the tests

tailscale/resource_tailnet_key.go
michaelbeaumont: Nice catch, done

davidsbond (approved):
LGTM

PR #72, opened 2022-02-23
Add wait_for argument to tailscale_device data source

### LINKED ISSUE #69: FR: add retry and timeout options to the tailscale_device data source

My end goal is to be able to define EC2 instances in Terraform and setup DNS entries for the tailscale IP address.  The current problem I have is that, unlike an AWS public or private IP, the tailscale IP address is unknown to AWS and not returned by their APIs.  Furthermore, there is a delay between the time the instance is created and it's user_data script can run  `tailscale up ...`.  So, using the `tailscale_device` data source in the same script usually results in that API call failing.

I'd like to be able to do something like:

```hcl
resource "aws_instance" "enterprise" {
    # ...snip...
    user_data = templatefile("tailscale-init.sh", {
        hostname = "enterprise"
        tailscale_auth_key = '...snip...'
    })
}


resource "aws_route53_record" "enterprise_pub" {
    zone_id = var.some_zone_id
    name = "enterprise"
    type = "A"
    ttl = "300"
    records = [
        aws_instance.enterprise.public_ip
    ]
}


data "tailscale_device" "enterprise" {
  name = "enterprise.${var.some_zone_domain}"
  # Retry API call 10 times waiting 10 seconds in-between each try
  retry_count = 10
  retry_interval = "10s"
}


resource "aws_route53_record" "enterprise_ts" {
    zone_id = var.some_zone_id
    name = "enterprise.ts"
    type = "A"
    ttl = "300"
    records = [
        # Another FR: should have .addresses_ipv4?
        tailscale_device.enterprise.addresses[0]
    ]
}
```

**Additional context**

If the above doesn't seem like the right approach, I'd be open to other ideas about how to provision an EC2 instance and get DNS setup for the Tailscale IP.

davidsbond:
This commit adds a new argument to the `tailscale_device` data source that allows for the configuration
of a retry mechanism. If this argument is provided, the provider will retry obtaining device data every
second up until the maximum duration specified in the `wait_for` argument.

This argument is optional, if it is not specified then the first error will fail the read. The field
is expected to use the string representation of go's `time.Duration` type.

Closes #69 (nice)

### DIFF
--- a/docs/data-sources/device.md
+++ b/docs/data-sources/device.md
@@ -14,6 +14,7 @@ The device data source describes a single device in a tailnet.
 ```terraform
 data "tailscale_device" "sample_device" {
   name = "user1-device.example.com"
+  wait_for = "60s"
 }
 
 ```
@@ -21,6 +22,7 @@ data "tailscale_device" "sample_device" {
 ## Argument Reference
 
 - `name` - (Required) The name of the tailnet device to obtain the attributes of.
+- `wait_for` - (Optional) If specified, the provider will retry obtaining the device data every second until the specified duration has been reached. Must be a value greater than 1 second
 
 ## Attributes Reference
 
--- a/tailscale/data_source_device.go
+++ b/tailscale/data_source_device.go
@@ -2,8 +2,10 @@ package tailscale
 
 import (
 	"context"
+	"time"
 
 	"github.com/davidsbond/tailscale-client-go/tailscale"
+	"github.com/hashicorp/go-cty/cty"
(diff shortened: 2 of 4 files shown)


rsyring:
@davidsbond thanks for putting the effort into this PR and discussion on #69.  I can't really comment on the code as I'm not a Go developer, but from a general "once over" everything seems reasonable to me and I think this will result in the functionality I'm looking for.

davidsbond:
Thanks @rsyring, I'll merge this in now then and publish a release which should also include the tags functionality you were after. It can take ~20 minutes after publishing a release for it to be available in the Terraform registry.

PR #80, opened 2022-03-07
[resource] tailnet_key: remove the duplicate "id" field

stack72:
The Terraform Plugin SDK already has this functionality when you
call d.SetID

```
// SetId sets the ID of the resource. If the value is blank, then the
// resource is destroyed.
func (d *ResourceData) SetId(v string) {
	d.once.Do(d.init)
	d.newState.ID = v

	// once we transition away from the legacy state types, "id" will no longer
	// be a special field, and will become a normal attribute.
	// set the attribute normally
	d.setWriter.unsafeWriteField("id", v)

	// Make sure the newState is also set, otherwise the old value
	// may get precedence.
	if d.newState.Attributes == nil {
		d.newState.Attributes = map[string]string{}
	}
	d.newState.Attributes["id"] = v
}
```

Fixes #


davidsbond:
Hey @stack72, thanks for the contribution. Looks like a further change is needed here:

https://github.com/davidsbond/terraform-provider-tailscale/pull/80/files#diff-dfdde71a78d094cc50323a462f992c94ea0a9ac636a515e75817a307d0293c90R95

davidsbond:
Thanks, I'll publish a patch release

stack72:
You rock @davidsbond!!

PR #98, opened 2022-04-22
Allow declaration of tailscale api endpoint

kelcya:
Allow users to declare the baseurl in the provider block. Add a new baseurl variable.

fixes https://github.com/davidsbond/terraform-provider-tailscale/issues/97

### DIFF
--- a/tailscale/provider.go
+++ b/tailscale/provider.go
@@ -34,7 +34,7 @@ func Provider(options ...ProviderOption) *schema.Provider {
 			"base_url": {
 				Type:        schema.TypeString,
 				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_BASE_URL", "https://api.tailscale.com"),
-				Required:    true,
+				Optional:    true,
 			},
 		},
 		ResourcesMap: map[string]*schema.Resource{
--- a/tailscale/provider.go
+++ b/tailscale/provider.go
@@ -31,9 +31,9 @@ func Provider(options ...ProviderOption) *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_TAILNET", nil),
 				Required:    true,
 			},
-			"baseurl": {
+			"base_url": {
 				Type:        schema.TypeString,
-				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_BASEURL", nil),
+				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_BASE_URL", "https://api.tailscale.com"),
 				Required:    true,
 			},
 		},
@@ -64,9 +64,9 @@ func Provider(options ...ProviderOption) *schema.Provider {
 func providerConfigure(_ context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
 	apiKey := d.Get("api_key").(string)
 	tailnet := d.Get("tailnet").(string)
-	baseurl := d.Get("baseurl").(string)
-	newbase := tailscale.WithBaseURL(baseurl)
-	client, err := tailscale.NewClient(apiKey, tailnet, newbase)
+	baseURL := d.Get("base_url").(string)
+
+	client, err := tailscale.NewClient(apiKey, tailnet, tailscale.WithBaseURL(baseURL))
 	if err != nil {
 		return nil, diagnosticsError(err, "failed to initialise client")
 	}
--- a/tailscale/provider.go
+++ b/tailscale/provider.go
@@ -31,6 +31,11 @@ func Provider(options ...ProviderOption) *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_TAILNET", nil),
 				Required:    true,
 			},
+			"baseurl": {
+				Type:        schema.TypeString,
+				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_BASEURL", nil),
+				Required:    true,
+			},
 		},
 		ResourcesMap: map[string]*schema.Resource{
 			"tailscale_acl":                  resourceACL(),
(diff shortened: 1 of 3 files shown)


tailscale/provider.go
```diff
@@ -31,6 +31,11 @@ func Provider(options ...ProviderOption) *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_TAILNET", nil),
 				Required:    true,
 			},
+			"baseurl": {
```
davidsbond, line 34:
```suggestion
			"base_url": {
```

tailscale/provider.go
```diff
@@ -31,6 +31,11 @@ func Provider(options ...ProviderOption) *schema.Provider {
...
+				Type:        schema.TypeString,
+				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_BASEURL", nil),
+				Required:    true,
```
davidsbond, line 37: I think this should be optional, and the default use `https://api.tailscale.com`

tailscale/provider.go
```diff
@@ -31,6 +31,11 @@ func Provider(options ...ProviderOption) *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_TAILNET", nil),
 				Required:    true,
 			},
+			"baseurl": {
+				Type:        schema.TypeString,
+				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_BASEURL", nil),
```
davidsbond, line 36:
```suggestion
				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_BASE_URL", nil),
```

tailscale/provider.go
```diff
@@ -59,8 +64,9 @@ func Provider(options ...ProviderOption) *schema.Provider {
 func providerConfigure(_ context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
 	apiKey := d.Get("api_key").(string)
 	tailnet := d.Get("tailnet").(string)
-
-	client, err := tailscale.NewClient(apiKey, tailnet)
+	baseurl := d.Get("baseurl").(string)
```
davidsbond, line 67:
```suggestion
	baseurl := d.Get("base_url").(string)
```

tailscale/provider.go
```diff
@@ -59,8 +64,9 @@ func Provider(options ...ProviderOption) *schema.Provider {
...
+	newbase := tailscale.WithBaseURL(baseurl)
+	client, err := tailscale.NewClient(apiKey, tailnet, newbase)
```
davidsbond, line 69:
```suggestion
	client, err := tailscale.NewClient(apiKey, tailnet, 
            tailscale.WithBaseURL(baseurl),
	)
```

davidsbond (commented):
Thanks for raising this, I've suggested a few changes.

I would also be good if you could update the documentation regarding the provider configuration values.

kelcya:
@davidsbond I added your suggestions, updated documentation and added the default api endpoint so `base_url` is optional. Tested locally and works.

davidsbond:
Hey @kelcya, thanks for addressing the comments.

For the documentation, I should have been clearer. The file here: https://github.com/davidsbond/terraform-provider-tailscale/blob/master/docs/index.md is what ends up on the terraform registry. Could you please add some information on the new provider configuration values there too?

Thanks :)

kelcya:
> Hey @kelcya, thanks for addressing the comments.
> 
> For the documentation, I should have been clearer. The file here: https://github.com/davidsbond/terraform-provider-tailscale/blob/master/docs/index.md is what ends up on the terraform registry. Could you please add some information on the new provider configuration values there too?
> 
> Thanks :)

Good to know. Updated.

tailscale/provider.go
```diff
@@ -31,6 +31,11 @@ func Provider(options ...ProviderOption) *schema.Provider {
 				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_TAILNET", nil),
 				Required:    true,
 			},
+			"base_url": {
+				Type:        schema.TypeString,
+				DefaultFunc: schema.EnvDefaultFunc("TAILSCALE_BASE_URL", "https://api.tailscale.com"),
+				Required:    true,
```
davidsbond, line 37:
```suggestion
```

I don't know exactly what the behaviour will be if required is set to true and there's a default, so I don't think this should be a required field.

davidsbond (commented):
Thanks, one last thing to change and I think this can be merfed

tailscale/provider.go
kelcya: I have tried setting it to `false` in the past and removing it just now. Both yield the same result.

Either will prompt for the `base_url` if you do not specify in the provider block.

```
kyang@kyang terraform % terraform plan
provider.tailscale.base_url
  Enter a value: https://api.tailscale.com

╷
│ Error: InternalValidate
│
│   with provider["kyang.local/testing/tailscale"],
│   on <input-prompt> line 1:
│   (source code not available)
│
│ Internal validation of the provider failed! This is always a bug
│ with the provider itself, and not a user issue. Please report
│ this bug:
│
│ 1 error occurred:
│ 	* base_url: One of optional, required, or computed must be set
│
│
╵
```

Only setting it to require to true seems to work the way intended. If you do not specify the `base_url`, it will use the default.

davidsbond: There should also be an `Optional` field you can set to true I think. I think you have to be explicit if a field is optional or required. A value of `false` in the `Required` field on its own does not indicate to the SDK that it is optional

davidsbond: https://github.com/hashicorp/terraform-plugin-sdk/blob/v2.14.0/helper/schema/schema.go#L44

Here's that type in particular. The `Optional` field should be settable here

kelcya: Looks good. Tested locally.

davidsbond:
Perfect. Thanks for addressing all the comments @kelcya I'll just wait for the tests to run then merge. If that's all good I'll publish a new release. A new release can take ~20 minutes to show up in the terraform registry once the release process is done.

kelcya:
> Perfect. Thanks for addressing all the comments @kelcya I'll just wait for the tests to run then merge. If that's all good I'll publish a new release. A new release can take ~20 minutes to show up in the terraform registry once the release process is done.

My pleasure. First time contributing to a public github repo. Thank you for writing this provider. I was using it to hack around Tailscale's lack of SCIM with our IDP but they recently rolled out a SCIM beta feature. Either way, it is useful. 👍

davidsbond:
Congratulations on your first contribution! That's great to hear :)

PR #102, opened 2022-05-10
Avoid using deprecated hujson API

dsnet:
The hujson.Unmarshal function is deprecated.
Use hujson.Standardize instead.

The hujson package is evolving to be just a syntactic parser for HuJSON
and is moving away from semantically understanding HuJSON as Go data.


dsnet:
The tests will fail since https://github.com/davidsbond/tailscale-client-go/pull/17 needs to be submitted first.

davidsbond:
Thanks @dsnet, I'll get that dependency upgraded then we can merge this

dsnet:
I updated to the latest version of `tailscale-client-go` in this PR. Up to you if you want to upgrade that dependency seperately.

davidsbond:
Thanks, I think we need to do both in one PR as #103 fails the CI

PR #133, opened 2022-07-21
remove required provider arguments

### LINKED ISSUE #132: terraform validate reports api_key as required
`terraform validate` reports `api_key` as missing from provider config when using `TAILSCALE_API_KEY` from the environment (or in Terraform Cloud):

```
Error: Missing required argument

  on tailscale.tf line 1, in provider "tailscale":
   1: provider "tailscale" {

The argument "api_key" is required, but no definition was found.
```

Not sure how other providers handle this to make the validation happy, but my best guess is that it should not be marked as required and fail at runtime when no valid API key has been configured.

hollow:
fixes #132


hollow:
@davidsbond I think this should do it ... this should just fail the tailscale client requests, but I have to admit that I didn't test the full cycle in Terraform Cloud with an unreleased provider

davidsbond:
Thanks @hollow, this looks good so far, the only thing missing is adding checks to [this function](https://github.com/hollow/terraform-provider-tailscale/blob/ef9ad3102290a72f54f6949c96bb338c42177c91/tailscale/provider.go#L69) that return errors if the `api_key` or `tailnet` fields are blank.

hollow:
@davidsbond my assumption was that the tailscale client does (or should) return an error here which is already handled [here](https://github.com/hollow/terraform-provider-tailscale/blob/ef9ad3102290a72f54f6949c96bb338c42177c91/tailscale/provider.go#L75), but if you prefer explicit checks for these arguments, I'll add them!

hollow:
@davidsbond I have tested the changes with local state without any provider configuration and it fails with:

```
│ Error: Failed to fetch ACL
│ 
│   with tailscale_acl.default,
│   on tailscale.tf line 5, in resource "tailscale_acl" "default":
│    5: resource "tailscale_acl" "default" {
│ 
│ API token invalid (401)
```

Looks like a decent error message to me. WDYT?

davidsbond:
The tailscale client does not error when the api key or tailnet are missing, the API calls will just fail. This can be seen [here](https://github.com/davidsbond/tailscale-client-go/blob/master/tailscale/client.go#L51). 

My problem with the above is that it is not clear to the user of the provider that they've neglected to set the value in the provider. I think adding some specific error messages that state that the provider configuration is invalid will be a better experience overall for users.

hollow:
The error message looks like this now:

```
│ Error: tailscale provider argument `tailnet' is empty
│ 
│   with provider["github.com/davidsbond/tailscale"],
│   on tailscale.tf line 1, in provider "tailscale":
│    1: provider "tailscale" {
```

tailscale/provider.go
```diff
@@ -68,7 +68,15 @@ func Provider(options ...ProviderOption) *schema.Provider {
 
 func providerConfigure(_ context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
 	apiKey := d.Get("api_key").(string)
+	if apiKey == "" {
+		return nil, diag.Errorf("tailscale provider argument `api_key' is empty")
```
davidsbond, line 72:
```suggestion
		return nil, diag.Errorf("tailscale provider argument 'api_key' is empty")
```

tailscale/provider.go
```diff
@@ -68,7 +68,15 @@ func Provider(options ...ProviderOption) *schema.Provider {
...
+	}
+
 	tailnet := d.Get("tailnet").(string)
+	if tailnet == "" {
+		return nil, diag.Errorf("tailscale provider argument `tailnet' is empty")
```
davidsbond, line 77:
```suggestion
		return nil, diag.Errorf("tailscale provider argument 'tailnet' is empty")
```

davidsbond (commented):
Looks like a mix of backticks and apostrophes in the message, otherwise looks good

davidsbond:
This change has also modified the generated documentation. Please do the following:

1) `go install github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs` - To install the documentation generator
2) `go generate ./...` to update the documentation
3) Commit the result

hollow:
> Looks like a mix of backticks and apostrophes in the message, otherwise looks good

This is [Emacs Documentation Notation](https://www.emacswiki.org/emacs/EmacsDocumentationNotation) but either format is fine.

PR #160, opened 2022-10-10
resource_device_authorization: update delete

### LINKED ISSUE #68: FR: add ability to delete devices

When destroying resources with terraform, I'd like the ability to remove the device from Tailscale.  For example, when an AWS instance will be deleted, I want it's associated tailscale device to also be deleted.

Keep track of devices in Terraform state and, when the device is removed, use the [DELETE device](https://github.com/tailscale/tailscale/blob/main/api.md#device-delete) API to remove the device from the tailnet.

defo89:
deletes the device on authorization resouce deletion

**What this PR does / why we need it**: Ability to delete device with provider

**Which issue this PR fixes**: fixes #68

### DIFF
--- a/tailscale/resource_device_authorization.go
+++ b/tailscale/resource_device_authorization.go
@@ -113,6 +113,13 @@ func resourceDeviceAuthorizationUpdate(ctx context.Context, d *schema.ResourceDa
 }
 
 func resourceDeviceAuthorizationDelete(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-	// Since authorization cannot be removed at this point, deleting the resource will do nothing.
+	client := m.(*tailscale.Client)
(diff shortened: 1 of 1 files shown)


davidsbond (commented):
I'm still of the opinion that if the terraform provider is not in charge of creating devices then it seems like not the right place for deleting them. 

However, should this PR be accepted we should probably update the documentation for this resource as part of this pull request to inform users of the behaviour if they delete the `device_authorization` resource. 

IMO we should modify the description on line 14 to include this information, you'll also need to run `go generate ./...` after you modify it.

davidsbond:
> if multiple resources are created/destroyed in a single run, users may face an error that some resources failed to be destroyed (due to device being removed already).

This is part of the reason I don't like us managing deletion without creation, the UX suffers in a non-obvious way.

PR #163, opened 2022-10-12
Update sample provider config for new registry path

007:
The old versions are still published under `davidsbond`, the `tailscale` registry path only has `>= 0.13.5`. Copying config from documentation will not work.

Fixes #163


knyar:
I think the version has been updated in #171

PR #171, opened 2022-10-24
readme: add instructions to update

DentonGentry:
Need to `terraform state replace-provider`

Signed-off-by: Denton Gentry <dgentry@tailscale.com>

Explain how to update an existing installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ terraform {
   required_providers {
     tailscale = {
       source = "tailscale/tailscale"
-      version = "0.2.0"
+      version = "0.13.5"
     }
   }
 }
@@ -43,6 +43,12 @@ provider "tailscale" {
   tailnet = "example.com"
   base_url = "https://api.us.tailscale.com"
 }
+```
+
+## Updating an existing installation
+To update an existing terraform deployment currently using the original `davidsbond/tailscale` provider, use:
+```
+terraform state replace-provider registry.terraform.io/davidsbond/tailscale registry.terraform.io/tailscale/tailscale
 ```
 
 ## Contributing


README.md
```diff
@@ -45,6 +45,17 @@ provider "tailscale" {
...
 ```
 
+## Updating an existing installation
+To update an existing terraform deployment currently using the original `davidsbond/tailscale` provider, use:
+```
+terraform state replace-provider registry.terraform.io/davidsbond/tailscale registry.terraform.io/tailscale/tailscale
+```
+
+It might also be necessary to update the terraform\_version. We have one report that it was
+necessary to set `terraform_version="1.1.5"` (from an older version) before the tailscale/tailscale
```
knyar, line 55: I have read that report, but I am not really sure what the `terraform_version` referred to here actually is.

There is a [terraform_version parameter](https://learn.hashicorp.com/tutorials/terraform/versions?in=terraform/cli#inspect-the-terraform-state-file) in the state file that gets updated automatically when a newer version of Terraform is used to apply a given module. It should typically not be changed manually.

There is also a configuration parameter called [required_version](https://developer.hashicorp.com/terraform/language/settings#specifying-a-required-terraform-version) that we could recommend setting to the minimum version that the provider supports (something like `required_version = ">= 1.1.5"`). But we need to do more research to understand what that minimum version is. I think it's unlikely to be 1.1.5 (it's a patch version that has [just a few unrelated changes](https://github.com/hashicorp/terraform/blob/v1.1/CHANGELOG.md#115-february-02-2022) compared to 1.1.4).

I would probably suggest to omit this for now until we get clearer signal that this is necessary.

knyar (approved):

README.md
DentonGentry: Omitted

knyar (approved):

PR #186, opened 2022-12-01
Do not overwrite ACL file that has been changed manually

### LINKED ISSUE #182: tailscale_acl creation deletes already present ACL without any warning
**Describe the bug**
If a new `tailscale_acl` resource is getting created, it will delete and override an already existing ACL. This is the worst possible behaviour imaginable for a resource and can mess people up big time, especially since the Tailscale WebUI doesn't offer revision control of the ACL to restore previous ACLs. I would even suggest to pull this resource from circulation in an emergency release until this issue is fixed, since it is dangerous and doesn't heed the most basic concepts of a terraform resource one would expect. A "create" action should never ever override existing items.

So 2 things: If this resource continues to override the ACL wholesale then it should only do so after a successful (so far apparently unimplemented) import of the resource, and otherwise error. However, such a resource will be pretty much unusable for all but the most simple use cases, since usually all the Subnet Routers will live in different VPCs of different projects, and will be implemented individually as part of these projects. Resources aren't separated by provider in Terraform, but by organisational structures within a company.

Therefore the proper fix would be to make this resource merge its policy items into an already existing ACL.

**To Reproduce**
1. Create an ACL in tailscale WebUI manually
2. Create a different `tailscale_acl` via terraform, and apply it against the same tailnet.
3. The manually created ACL will get overridden

**Expected behaviour**
Never ever delete existing infrastructure/config without a lot of warning!

 - OS: Linux
 - Terraform Version 1.2.9
 - Provider Version 0.13.5

knyar:
Overwriting non-default ACL contents is confusing, so now users will be prompted to explicitly import their existing ACL into Terraform if they want to manage it there. This will make overwriting existing contents harder, since after importing Terraform will show a sensible diff in `terraform apply`.

Fixes #182.


fylie (approved):
awesome 😄

PR #203, opened 2023-02-14
Added `name_regexp` to tailscale_devices data source. 

### LINKED ISSUE #202: Please add an option to retrieve tailscale_devices (Data Source) via regexp
I want to add tags on tailscale devices based on their name. Unfortunately the data source `tailscale_devices` only has the option to match by name prefix. Our devices would be identified by the suffix.

Add more flexible option to match tailscale devices via `tailscale_devices` data source.

**Additional context**
Will add a PR to add `name_regexp` as the most flexible way to identify devices.

gitbensons:
This can be used to identify/match devices based on a regular expression RE2.

Adds option to search for devices using regular expression RE2.

Fixes #202 


knyar:
Thank you for proposing this change, and apologies that it took so long to review it.

I am not sure I understand why this is necessary. If filtering is happening on the client anyway, can't you use a `for` expression to select the devices you need as part of your Terraform configs? For example:

```
[for d in data.tailscale_devices.sample_devices.devices : d if length(regexall("-(mobile|laptop)$", d.name)) > 0]
```

PR #207, opened 2023-02-23
Update documentation

### LINKED ISSUE #206: Tailscale provider returns 403 when user tailnet does not match
One of the more common (I suspect) issues a new user of the provider may run into is that the tailnet name may not be correct. In this case the provider will return an error like this:

```
│ Error: Failed to create key
│
│   with module.my_module.tailscale_tailnet_key.default,
│   on ../modules/gcp_gitlab_runner/[main.tf](http://main.tf/) line 28, in resource "tailscale_tailnet_key" "default":
│   28: resource "tailscale_tailnet_key" "default" {
│
│ user tailnet does not match (403)
╵
```
I encountered this issue on the free tier and was able to resolve it by referring to the [API documentation](https://github.com/tailscale/tailscale/blob/main/api.md#tailnet). My tailnet name was the same as my organization name, i.e. equivalent to "example.com" in the API documentation. I'm not immediately sure how the tailnet name changes when you start paying for tailscale as the API docs are not all that clear.

I think that README.md could be updated to make this a little clearer.

heretic098:

Update the documentation to make it clearer how to handle tailnet names.

Fixes #206 

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,14 @@ provider "tailscale" {
 ```
 
 In the `provider` block, replace `api_key` and `tailnet` with your own tailnet and API key. Alternatively, use the
-`TAILSCALE_API_KEY` and `TAILSCALE_TAILNET` environment variables.
+`TAILSCALE_API_KEY` and `TAILSCALE_TAILNET` environment variables. The api_key can also take the Terraform file function, e.g if your API key is in a file called `creds/tailscale.key`:
+
+```
+provider "tailscale" {
+  api_key = file("../creds/tailscale.key")
+  tailnet = "example.com"
+}
+```
 
 The default api endpoint is `https://api.tailscale.com`. If your coordination/control server api is at another endpoint, you can pass in `base_url` in the provider block.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,7 +33,14 @@ provider "tailscale" {
 ```
 
 In the `provider` block, replace `api_key` and `tailnet` with your own tailnet and API key. Alternatively, use the
-`TAILSCALE_API_KEY` and `TAILSCALE_TAILNET` environment variables.
+`TAILSCALE_API_KEY` and `TAILSCALE_TAILNET` environment variables. The api_key can also take the Terraform file function, e.g if your API key is in a file called `creds/tailscale.key`:
+
+```
+provider "tailscale" {
+  api_key = file("../creds/tailscale.key")
+  tailnet = "example.com"
```
DentonGentry, line 41: I think this can be `tailnet = "-"` which means "the tailnet of the API key" and avoid the error about the tailnet name not matching.

PR #212, opened 2023-03-07
Add support for OAuth client credentials

### LINKED ISSUE #198: OAuth client for provider auth

API key that is currently required for provider configuration has to be rotated manually when expired. As an alternative - OAuth was recently introduced (in beta) for all tailnets: https://github.com/tailscale/tailscale/issues/3243#issuecomment-1405360347. 
In general OAuth client provides better security experience.

Be able to use the OAuth client for Tailscale provider auth instead of the API key.

clstokes:

Allows configuring the tailscale provider with OAuth Client ID and OAuth Client secret instead of a Tailscale API access token.

If configured with OAuth credentials, the tailscale provider will generate an API access token from the Tailscale OAuth token endpoint.

Fixes #198


clstokes:
It occurs to me this could and probably should be moved into https://github.com/tailscale/tailscale-client-go. I'll close this and look into that.

PR #221, opened 2023-03-23
Add `rotate_when_changed` to tailnet key schema

### LINKED ISSUE #144: The `tailscale_tailnet_key` resource should handle expired keys
**Describe the bug**
A few months after generating a few `tailscale_tailnet_key` resources in my terraform code, those keys expired in tailscale control and were subsequently deleted by the api. Now when I run my Terraform again after the keys have expired, the provider fails to fetch these keys and blocks my Terraform deployment.

**To Reproduce**
1. Terraform apply a `tailscale_tailnet_key` with `reusable = true`
2. Wait for the key to expire (or revoke it so you don't have to wait around)
3. Terraform plan/apply using same configuration
4. Fail to fetch key with 404
```
Error: Failed to fetch key

  with tailscale_tailnet_key.reusable,
  on my_keys.tf line 1, in resource "tailscale_tailnet_key" "reusable":
   1: resource "tailscale_tailnet_key" "reusable" {

404 page not found (404)
```

**Expected behaviour**
The provider detects an expired key by evaluating that the expire time/date in state is less than the current time/date and that the resource is a 404 response from the api, removes the resource from state, then a plan is calculated to recreate the resource to ensure the desired end state.

 - OS: linux
 - Terraform Version 1.1.9
 - Provider Version 0.12.2

**Additional context**
This is similar to an issue brought up in #59 where keys would vanish from the API, except due to `reusable = false`

omBratteng:
This adds a new field to the schema of `tailscale_tailnet_key`, which would allow one to use the [`time_rotating`](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/rotating) without depending on the lifecycle.

So you could do this to auto rotate the key

```hcl
resource "time_rotating" "autorotator" {
  rotation_days = 60
}

resource "tailscale_tailnet_key" "tailscale_ingress" {
  reusable  = true
  ephemeral = true
  tags = [
    "tag:some-tag",
  ]
  rotate_when_changed = {
    rotator = time_rotating.autorotator.id
  }
}
```

Fixes #144

### DIFF
--- a/tailscale/resource_tailnet_key.go
+++ b/tailscale/resource_tailnet_key.go
@@ -56,6 +56,15 @@ func resourceTailnetKey() *schema.Resource {
 				Description: "The expiry of the key in seconds",
 				ForceNew:    true,
 			},
+			"rotate_when_changed": {
+				Description: "Arbitrary map of values that, when changed, will trigger rotation of the key.",
+				Type:        schema.TypeMap,
+				Optional:    true,
+				ForceNew:    true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
 		},
 	}
 }


knyar:
Thanks for proposing this, and apologies that it took so long to review.

Seems a bit weird to add a bogus field just to trigger a recreation of the resource. Perhaps a viable alternative would be to add support for the `Description` field ([recently added](https://github.com/tailscale/tailscale-client-go/pull/53) to the API library)? This will allow you to use the autorotator ID as part of the description text, re-creating the key when the ID changes.

omBratteng:
The solution is inspired by the [`azuread_application_password` resource](https://github.com/hashicorp/terraform-provider-azuread/blob/df2d1be2e1943497e6d2ec7db144cb896caeb24e/internal/services/applications/application_password_resource.go#L93-L101), where they also have the `rotate_when_changed` field. Ideally it would've been a built-in terraform feature.

Kinda does feel a bit hacky to use the description field to trigger a rotation, and would be dependent on the user to add a rotation ID to the description field. Personally I would prefer a dedicated field for it.

knyar:
Yeah, I agree that this seems like something that Terraform should natively support. Adding a separate field that does not impact resource configuration at all does not seem like a good long term solution, since if/when Terraform does provide first-class support for this, that field would not longer be helpful at all. With this in mind, I'd prefer using an existing field as a work-around.

Key description support has been added in #277, so I'd recommend using that.

PR #238, opened 2023-05-22
docs: clarify tailnet var

### LINKED ISSUE #206: Tailscale provider returns 403 when user tailnet does not match

santiagon610:
**What this PR does / why we need it**: This PR clarifies some of the documentation related to the `tailnet` provider variable. Since it took digging into the issues (see below for references) to find the appropriate value, this may help future users.

- Fixes #206 
- Mentions #207


whosthegoss (approved):

whosthegoss:
LGTM

santiagon610:
Thanks @davidsbond, since I don't have any insight into the CI jobs, I was able to perform the steps that you recommended. Since `tfplugindocs validate .` gave me an exit code of 0, I assume I _should_ be good now.

```
❯ tfplugindocs validate .
detected static docs directory, running checks
❯ echo $?
0
```

davidsbond:
Yep that should be fine! Someone on the tailscale team should be able to run the CI here, approve & merge for you. I can't do that.

bailey-ts:
Hey @DentonGentry, would it be possible to get a review on this?

knyar (approved):

PR #277, opened 2023-08-25
Add tailscale tailnet key Description and update docs

julianorchard:

Adding the ability to define a description when creating a key. I've included an example, updated the docs with tfplugindocs, and added `description` to test.

Fixes #

Following on from addition in [the API](https://github.com/tailscale/tailscale-client-go/pull/53).

### DIFF
--- a/tailscale/resource_tailnet_key.go
+++ b/tailscale/resource_tailnet_key.go
@@ -66,6 +66,12 @@ func resourceTailnetKey() *schema.Resource {
 				Description: "The expiry timestamp of the key in RFC3339 format",
 				Computed:    true,
 			},
+			"description": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "A description of the key consisting of alphanumeric characters.",
+				ForceNew:    true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


tailscale/resource_tailnet_key.go
```diff
@@ -66,6 +66,12 @@ func resourceTailnetKey() *schema.Resource {
 				Description: "The expiry timestamp of the key in RFC3339 format",
 				Computed:    true,
 			},
+			"description": {
```
knyar, line 69: Could you please also set this field in `resourceTailnetKeyRead`?

tailscale/resource_tailnet_key.go
```diff
@@ -76,6 +82,7 @@ func resourceTailnetKeyCreate(ctx context.Context, d *schema.ResourceData, m int
 	ephemeral := d.Get("ephemeral").(bool)
 	preauthorized := d.Get("preauthorized").(bool)
 	expiry, hasExpiry := d.GetOk("expiry")
+	description := d.Get("description").(string)
```
knyar, line 85: Perhaps use a GetOk here, with a separate hasDescription bool that can be used as part of the condition below (instead of len)? This will make it more consistent with how we handle `expiry`.

knyar (commented):
Thank you, just a couple small comments.

julianorchard:
Thanks for the feedback, please let me know if there's anything else I need to change following my updated push! :)

PR #279, opened 2023-09-01
docs: Document how to configure an exit node

mhutter:
This PR documents how to automatically approve a device to be used as an exit node.

Even though it makes sense, it took quite a bit of figuring-out to do, so hopefully this saves someone else some time.

PS: As someone who never developed their own Terraform plugin, the contribution docs could also need some love, or at least a Makefile target on how to update the docs after a change :)


docs/resources/device_subnet_routes.md
```diff
@@ -22,7 +22,10 @@ resource "tailscale_device_subnet_routes" "sample_routes" {
   routes = [
     "10.0.1.0/24",
     "1.2.0.0/16",
-    "2.0.0.0/24"
+    "2.0.0.0/24",
+    # Or configure as an exit node
+    "0.0.0.0/0",
+    "::/0"
```
knyar, line 28: This makes this example unrealistic, since it does not quite make sense to announce 0/0 alongside more specific subnets.

May I suggest to add another `tailscale_device_subnet_routes` block here that would just have two 0/0 routes, providing an example of an exit node?

mhutter: Good point, I have adjusted the example accordingly

knyar (approved):
Thank you!

PR #303, opened 2023-11-15
added `allow_overwrite` option to resource_acl

### LINKED ISSUE #229: Failed to set ACL
Terraform version: 1.3.5
Tailscale provider version: 0.13.7
OS: Ubuntu 20.04

I am trying to set a Tailscale ACL via the provider for the first time.  There is no existing state for the ACL and I expect the Terraform state to be generated as part of the first apply, however, I am encountering errors requesting that I import the ACL?  Please advise.

```
tailscale_acl.default: Creating...
2023-04-21T12:00:31.337-0700 [INFO]  Starting apply for tailscale_acl.default
2023-04-21T12:00:31.338-0700 [DEBUG] tailscale_acl.default: applying the planned Create change
2023-04-21T12:00:31.934-0700 [ERROR] provider.terraform-provider-tailscale_v0.13.7: Response contains error diagnostic: @caller=github.com/hashicorp/terraform-plugin-go@v0.14.3/tfprotov5/internal/diag/diagnostics.go:55 diagnostic_summary="Failed to set ACL" tf_proto_version=5.3 tf_resource_type=tailscale_acl tf_rpc=ApplyResourceChange tf_req_id=599122e0-8509-7d08-1127-f1d752314276 @module=sdk.proto diagnostic_detail="! You seem to be trying to overwrite a non-default ACL with a tailscale_acl resource.
Before doing this, please import your existing ACL into Terraform state using:
 terraform import $(this_resource) acl
(got error "precondition failed, invalid old hash (412)")" diagnostic_severity=ERROR tf_provider_addr=provider timestamp=2023-04-21T12:00:31.934-0700
2023-04-21T12:00:31.934-0700 [ERROR] vertex "tailscale_acl.default" error: Failed to set ACL
╷
│ Error: Failed to set ACL
│
│   with tailscale_acl.default,
│   on acls.tf line 1, in resource "tailscale_acl" "default":
│    1: resource "tailscale_acl" "default" {
│
│ ! You seem to be trying to overwrite a non-default ACL with a tailscale_acl
│ resource.
│ Before doing this, please import your existing ACL into Terraform state
│ using:
│  terraform import $(this_resource) acl
│ (got error "precondition failed, invalid old hash (412)")
```

markwellis:
so resource_acl doesn't need to be imported first, as this is a manual task that breaks our workflow.

by default it is false, so will require import unless you read the docs and learn about the `allow_overwite` option

Fixes #229

It's my first time writing go, sorry if it's not best practice.


knyar:
Thank you for preparing a PR for this. Before adding more options, I'd like to understand what problem this solves and make sure this is something that other users will find helpful as well. Could you please share a bit more on why this is necessary?

`resource_acl` should support `terraform import`, so if you have a non-default policy, you should be able to import it before changing. For tailnets that never had their ACL changed from the default, this check should pass and allow setting the policy even without importing.

markwellis:
Yes it is a little strange, but the way we do it is to manage parts of the ACL in places, by loading the existing ACL, and then updating certain groups etc. It is riddled with race conditions, but it works better for us than having one place that manages the ACL.

What would be best would be if the tailscale API allowed us to mange groups/acls/hosts etc individually, rather than as a single json document. But there's nothing I can do about that

E.g. what we do is add a host for a new VPC's network, a new group with the VPC name in, and an ACL that allows the group access to the host. Doing that when we create the VPC is neater for us than doing it in a single place. It means the ACL for accessing the VPC lives with the VPC terraform itself, rather than somewhere else. Even though this isn't how the tailscale API actually works

it would be awesome if we could do something like
```terraform
resource "aws_vpc" "example" {
  ...
}

resource "tailscale_acl_host" "foo" {
  name = aws_vpc.example.tags.name
  ip_address = aws_vpc.example.cidr_block
}

resource "tailscale_acl_group" "foo"  {
  name = aws_vpc.example.tags.name
  members = [] #managed in the tailscale admin or scim in the future
}

resource "tailscale_acl_acl" "foo" {
  action = "accept"
  src = tailscale_acl_group.foo.name
  dst = tailscale_acl_host.foo.name
}
```

but we can't so instead we do something like

```terraform
resource "aws_vpc" "example" {
  ...
}

data "tailscale_acl" "acl" {
}

locals {
  acl = jsondecode(data.tailscale_acl.acl.json)
}

import {
  to = tailscale_acl.acl
  id = "acl"
}
resource "tailscale_acl" "acl" {
  acl = jsonencode(merge(local.acl, {groups = merge(local.acl.groups, {"hosts": {aws_vpc.example.tags.name
 = aws_vpc.example.cidr_block}, "group:foo" = []}, ...)}))
}
```
where we abuse merge to overwrite just parts of the json. it works, but it's not great, but is still better for us than managing the acl in one place

hope this clears up our ~abuse~use case

markwellis:
I'm going to close this as i can use the `import` block as a work around

markwellis:
reopening as import blocks cannot be used in modules, see https://github.com/hashicorp/terraform/issues/33474

Because of that issue I would still like to be able to set `allow_overwrite` on the `tailscale_acl` resource to not require importing it before terraform can manage the ACL

tailscale/resource_acl.go
```diff
@@ -95,7 +100,12 @@ func resourceACLCreate(ctx context.Context, d *schema.ResourceData, m interface{
 
 	// Setting the `ts-default` ETag will make this operation succeed only if
 	// ACL contents has never been changed from its default value.
-	if err := client.SetACL(ctx, acl, tailscale.WithETag("ts-default")); err != nil {
+	var opts []tailscale.SetACLOption
+	if d.Get("allow_overwrite") != true {
```
knyar, line 104: Perhaps
```suggestion
	if !d.Get("allow_overwrite").(bool) {
```

tailscale/resource_acl.go
```diff
@@ -36,6 +36,11 @@ func resourceACL() *schema.Resource {
 				ValidateDiagFunc: validateACL,
 				Description:      "The JSON-based policy that defines which devices and users are allowed to connect in your network",
 			},
+			"allow_overwrite": {
```
knyar, line 39: This seems a bit too ambiguous of a name. For example, it's not clear right now that this only applies when a new resource is created.

Any thoughts on using a more specific name? For example, `overwrite_on_creation` or `overwrite_existing_content`?

knyar (commented):
Thank you for your patience here. Just a couple comments from me.

tailscale/resource_acl.go
markwellis: hmm, good point. I just used the one I'm familiar with, from  https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_record#allow_overwrite.

I think `overwrite_existing_content` is clearer than `overwrite_on_creation`

knyar: If you could please rename the attribute to `overwrite_existing_content` I will be happy to merge and cut a new release. Thank you!

markwellis: updated, thanks

PR #287, opened 2023-09-12
resource_tailnet_key: handle 404 and discard invalid keys

clstokes:
**What this PR does / why we need it**:

1. Changes the handling of 404 errors on Read / `terraform refresh` operations to treat all 404 errors as a deleted key. Previously only one-time keys (e.g _not_ reusable) were treated as deleted and a 404 would fail the terraform operation with an error.
2. Removes tailnet key from state during Read operation if `invalid` key property is `true`.

**Which issue this PR fixes**: 

Fixes #144

**Special notes for your reviewer**:

~Any concerns with using the time of the machine that is running Terraform to compare with `expires_at` returned by the Tailscale API?~

PR #310, opened 2023-12-08
tailscale_tailnet_key: only recreate reusable keys by default

### LINKED ISSUE #306: Don't delete `tailscale_tailnet_key` after the key is used.
**Describe the bug**
It doesn't seem possible to use `tailscale_tailnet_key` in practice, because as soon as it's used once, the key is "deleted" and Terraform wants to replace the resource, which causes a cascade of updates/replacements, when actually the key may not be needed anymore/it's saved in the state file.

```
lifecycle {
  ignore_changes = all
}
```

doesn't seem to work. Is there a feature of this resource/Terraform that I might be missing?

**To Reproduce**
Use the `tailscale_tailnet_key` resource to trigger creating a new machine and then run `terraform apply` again.

**Expected behaviour**
I should be able to use `tailscale_tailnet_key` and not have it trigger a recreation.

knyar:
This change partially reverts the behaviour introduced in #287 that currently results in single-use keys being recreated, triggering unnecessary updates to downstream Terraform resources.

By default, the provider will now only recreate reusable keys, ignoring invalid single-use keys. This can also be changed now using a new `recreate_if_invalid` attribute.

Fixes #306

### DIFF
--- a/docs/resources/tailnet_key.md
+++ b/docs/resources/tailnet_key.md
@@ -31,6 +31,7 @@ resource "tailscale_tailnet_key" "sample_key" {
 - `ephemeral` (Boolean) Indicates if the key is ephemeral. Defaults to `false`.
 - `expiry` (Number) The expiry of the key in seconds. Defaults to `7776000` (90 days).
 - `preauthorized` (Boolean) Determines whether or not the machines authenticated by the key will be authorized for the tailnet by default. Defaults to `false`.
+- `recreate_if_invalid` (String) Determines whether the key should be created again if it becomes invalid. By default, reusable keys will be recreated, but single-use keys will not. Possible values: 'always', 'never'.
 - `reusable` (Boolean) Indicates if the key is reusable or single-use. Defaults to `false`.
 - `tags` (Set of String) List of tags to apply to the machines authenticated by the key.
 
@@ -39,4 +40,5 @@ resource "tailscale_tailnet_key" "sample_key" {
 - `created_at` (String) The creation timestamp of the key in RFC3339 format
 - `expires_at` (String) The expiry timestamp of the key in RFC3339 format
 - `id` (String) The ID of this resource.
+- `invalid` (Boolean) Indicates whether the key is invalid (e.g. expired, revoked or has been deleted).
 - `key` (String, Sensitive) The authentication key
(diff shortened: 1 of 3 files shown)


clstokes:
I ran through a few scenarios with this branch and everything looked good except for one update-in-place scenario. 

If I change `recreate_if_invalid = "never" -> "always"` on an expired and invalid key and then `apply`, the property gets changed on the first apply and the next subsequent `apply` replaces the key. I think the replace should happen in the first apply.

#### Steps to reproduce

1. `git clone git@gist.github.com:4adf7c5b90e60c2cfaa632f027b4132f.git` ([gist url](https://gist.github.com/clstokes/4adf7c5b90e60c2cfaa632f027b4132f))
2. `terraform apply -auto-approve`
3. `terraform output key` _<--- Observe `invalid = false`_
4. `sleep 30`
5. `echo 'recreate_if_invalid = "always"' > terraform.tfvars`
6. `terraform apply -auto-approve` _<--- Observe `recreate_if_invalid` is updated, but key is not replaced._
3. `terraform output key` _<--- Observe `invalid = true`_
7. `terraform apply -auto-approve`  _<--- Observe Key is replaced._

#### Output

[see gist](https://gist.github.com/clstokes/4adf7c5b90e60c2cfaa632f027b4132f#file-z-output)

knyar:
> If I change recreate_if_invalid = "never" -> "always" on an expired and invalid key and then apply, the property gets changed on the first apply and the next subsequent apply replaces the key. I think the replace should happen in the first apply.

Nice, thanks a lot for testing and for sharing this. This is not trivial to fix: the stage at which Terraform shows a simple diff happens before we have an opportunity to read the resource by default.

I have addressed this by adding a custom diff function for the resource that now triggers recreation if a flag gets changed. It duplicates some logic of the Read function that similarly recreates the key when it expires, but I don't think there's an elegant single place we could do this in. PTAL?

clstokes:
> Nice, thanks a lot for testing and for sharing this. This is not trivial to fix: the stage at which Terraform shows a simple diff happens before we have an opportunity to read the resource.
> 
> I have addressed this by adding a custom diff function for the resource that now triggers recreation if a flag gets changed. It duplicates some logic of the Read function that similarly recreates the key when it expires, but I don't think there's an elegant single place we could do this in. PTAL?

I ran through the same scenario and others and this looks good to me!

clstokes (approved):

michaelbeaumont:
@knyar Great, thanks!

PR #161, opened 2022-10-10
Perform validation on ACL create/update

davidsbond:
**What this PR does / why we need it**:

This commit modifies the provider to validate ACLs against the Tailscale API on create or update. Validation will only be performed if the ACL definition has at least one test in it.

Signed-off-by: David Bond <davidsbond93@gmail.com>

**Which issue this PR fixes** *(use `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: 

Fixes #158

PR #325, opened 2024-01-28
Document that ACL tests are run when creating and updating tailscale_acl resources.

### LINKED ISSUE #158: allowing validating the ACL tests before `terraform apply`
We prefer to manage our infrastructure with Terraform whenever possible. But one nice feature that [Tailscale's GitOps Pusher](https://github.com/tailscale/tailscale/tree/main/cmd/gitops-pusher) supports is to validate your ACL logic pre-merge in a PR by asking Tailscale to run the ACL tests. It'd be great if the Terraform provider could support this somehow as well.

I'm not entirely sure how to best fit this into a Terraform workflow -- my first thought would be to have `terraform plan` run the validation step. But that would perhaps be surprising to some?

dsymonds:
Fixes #158.
Closes #161.

### DIFF
--- a/docs/resources/acl.md
+++ b/docs/resources/acl.md
@@ -4,11 +4,14 @@ page_title: "tailscale_acl Resource - terraform-provider-tailscale"
 subcategory: ""
 description: |-
   The acl resource allows you to configure a Tailscale ACL. See https://tailscale.com/kb/1018/acls for more information. Note that this resource will completely overwrite existing ACL contents for a given tailnet.
+  If tests are defined in the ACL (the top-level "tests" section), ACL validation will occur before creation and update operations are applied.
 ---
 
 # tailscale_acl (Resource)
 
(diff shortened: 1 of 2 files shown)


dsymonds:
@clstokes @davidsbond @knyar 

I've tested this manually, but not sure how easy or useful it is to add a unit test.

knyar:
Thank you! I might be missing something, but what does this achieve if validation is only called immediately before attempting to set ACL contents with the same value? The same validation (including tests) will be applied while ACL content is [being set](https://github.com/tailscale/tailscale/blob/main/api.md#update-policy-file), and an error should be returned if any of the tests fail.

dsymonds:
That's a good point! For some reason I thought this wasn't happening, but just tested it manually now and the tests do indeed get exercised.

Okay, let me pare this right back to just the documentation update.

dsymonds:
PTAL @knyar

dsymonds:
Ah ha! I overlooked the very subtle "generated" comment at the top. Done properly now.

PTAL @knyar

knyar (approved):
Thank you!

PR #359, opened 2024-04-24
tailscale: add `dns_split_nameservers` resource

mpminardi:
Add `resource_dns_split_nameservers` to allow for controlling split DNS settings for a given tailnet.

Updates https://github.com/tailscale/corp/issues/19483


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.33.0
 	github.com/stretchr/testify v1.9.0
 	github.com/tailscale/hujson v0.0.0-20221223112325-20486734a56a
-	github.com/tailscale/tailscale-client-go v1.16.0
+	github.com/tailscale/tailscale-client-go v1.16.1-0.20240424180809-5a248d9547d4
```
mpminardi, line 12: TODO:
- [ ] uprev to actual version once https://github.com/tailscale/tailscale-client-go/pull/78 is through and a new release is cut

docs/resources/dns_split_nameservers.md
```diff
@@ -0,0 +1,33 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "tailscale_dns_split_nameservers Resource - terraform-provider-tailscale"
+subcategory: ""
+description: |-
+  The dnssplitnameservers resource allows you to configure split DNS nameservers for your Tailscale network. See https://tailscale.com/kb/1054/dns for more information.
```
mpminardi, line 6: Tried to snake case this to match the actual resource name but the autogen tool / CI didn't like that so 🤷

tailscale/resource_dns_split_nameservers.go
```diff
@@ -0,0 +1,112 @@
...
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/tailscale/tailscale-client-go/tailscale"
+)
+
+func resourceDNSSplitNameservers() *schema.Resource {
+	return &schema.Resource{
```
knyar, line 13: I'd suggest adding an importer as well to allow users to import existing Split DNS configuration to their Terraform state. I suspect the usual `&schema.ResourceImporter{StateContext: schema.ImportStatePassthroughContext}` would work here.

tailscale/resource_dns_split_nameservers.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func resourceSplitDNSNameserversRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	client := m.(*tailscale.Client)
+	splitDNS, err := client.DNSSplitDNS(ctx)
+	if err != nil {
+		return diagnosticsError(err, "Failed to fetch split DNS configs")
+	}
+
+	nameservers := splitDNS[d.Get("domain").(string)]
```
knyar, line 44: We might need something like this to consider the resource deleted when there are no split DNS servers configured for a given domain.

```suggestion
	nameservers, ok := splitDNS[d.Get("domain").(string)]
	if !ok {
	  d.SetId("")
	  return
	}
```

tailscale/resource_dns_split_nameservers.go
```diff
@@ -0,0 +1,112 @@
...
+	d.SetId(domain)
+	return nil
+}
+
+func resourceSplitDNSNameserversUpdate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	if !d.HasChange("nameservers") {
+		return resourceSplitDNSNameserversRead(ctx, d, m)
+	}
+
+	client := m.(*tailscale.Client)
```
knyar, line 80: Would it make sense to just call `resourceSplitDNSNameserversCreate` here?

mpminardi: Good call, will just delegate off to create.

tailscale/resource_dns_split_nameservers.go
mpminardi: Digging into this a bit more but making this change was breaking import.

mpminardi: Ended up changing to use `d.Id()` as this is what was breaking the import and is mentioned as necessary [here](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/import#importer-state-function) if using `schema.ImportStatePassthrough`.

This seems to operate fine without the `ok` check in the scenarios I've tested (namely creation, deletion, import, update of split DNS nameservers) so I've left it out for now.

knyar (approved):

aiell0:
Just chiming in to say I love using Tailscale and this is a feature I have been anticipating for a while! Good work @mpminardi

PR #361, opened 2024-04-30
tailscale: fix default User-Agent header to include terraform version

mpminardi:
The `schema.Provider.UserAgent` helper function was previously being called at a point in the plugin lifecycle before the `TerraformVersion` for the provider was set by Terraform Core. This was resulting in the Terraform version information being missing from the User-Agent header. The call to `schema.Provider.UserAgent` has now been moved into the `providerConfigure` func as this is called later in the configure lifecycle where the Terraform version is properly set.

### Header Before
`Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.33.0 terraform-provider-tailscale/dev`

### Header After
`Terraform/1.8.2 (+https://www.terraform.io) Terraform-Plugin-SDK/2.33.0 terraform-provider-tailscale/dev`

Fixes https://github.com/tailscale/corp/issues/19630

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -5,7 +5,7 @@ builds:
     flags:
       - -trimpath
     ldflags:
-      - "-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}"
+      - "-s -w -X github.com/tailscale/terraform-provider-tailscale/tailscale.providerVersion={{.Version}} -X main.commit={{.Commit}}"
     goos:
       - freebsd
       - windows
--- a/main.go
+++ b/main.go
@@ -8,24 +8,10 @@ import (
 	"github.com/tailscale/terraform-provider-tailscale/tailscale"
 )
 
-// version is filled by goreleaser at build time.
-var version = "dev"
-
 func main() {
 	plugin.Serve(&plugin.ServeOpts{
 		ProviderFunc: func() *schema.Provider {
-			return tailscale.Provider(addUserAgent)
+			return tailscale.Provider()
 		},
 	})
-}
-
-// addUserAgent adds a `user_agent` configuration key to the provider with a
-// default value based on provider version.
-func addUserAgent(p *schema.Provider) {
-	p.Schema["user_agent"] = &schema.Schema{
-		Type:        schema.TypeString,
-		Default:     p.UserAgent("terraform-provider-tailscale", version),
-		Optional:    true,
-		Description: "User-Agent header for API requests.",
-	}
 }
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -8,24 +8,10 @@ import (
 	"github.com/tailscale/terraform-provider-tailscale/tailscale"
 )
 
-// version is filled by goreleaser at build time.
-var version = "dev"
```
mpminardi, line 12: Moved this to `provider.go` as that is the one place it is used currently and to avoid messiness with having to pass the version into the `Provider` method.

tailscale/provider.go
```diff
@@ -84,14 +91,18 @@ func Provider(options ...ProviderOption) *schema.Provider {
 		},
 	}
 
+	provider.ConfigureContextFunc = func(ctx context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
```
mpminardi, line 94: Moved setting `ConfigureContextFunc` here so we can pass the provider into `providerConfigure`.

tailscale/provider.go
```diff
@@ -113,6 +124,9 @@ func providerConfigure(_ context.Context, d *schema.ResourceData) (interface{},
 	}
 
 	userAgent := d.Get("user_agent").(string)
+	if userAgent == "" {
+		userAgent = provider.UserAgent("terraform-provider-tailscale", providerVersion)
```
mpminardi, line 128: This is essentially an identical call to what we were making before, but at this point the underlying `provider.TerraformVersion` value used in the [UserAgent call](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/provider.go#L497) is populated properly.

.goreleaser.yml
```diff
@@ -5,7 +5,7 @@ builds:
     flags:
       - -trimpath
     ldflags:
-      - "-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}"
+      - "-s -w -X github.com/tailscale/terraform-provider-tailscale/tailscale.providerVersion={{.Version}} -X main.commit={{.Commit}}"
```
mpminardi, line 8: This needs to be the fully qualified path since it no longer lives in `main.go`. See [here](https://github.com/hashicorp/terraform-provider-azurerm/blob/75d6f38bd5c92cdd4825d5a5324b5dfc26cc7104/.goreleaser.yml#L38) in the AzureRM Terraform provider for similar usage.

knyar (approved):

soniaappasamy (approved):
Nice! Ty for digging into this

PR #362, opened 2024-05-01
tailscale: update state after modification functions for resources

mpminardi:
Add calls to appropriate `Read` functions at the end of `Create` and `Update` functions to follow recommended best practices from HashiCorp. This is done to prevent scenarios where Terraform could otherwise have stale state after resource creation or update (from default or computed values being populated server side on the object as an example) which can result in unexpected drift in subsequent plan / apply operations.

Fixes https://github.com/tailscale/corp/issues/19698

### DIFF
--- a/tailscale/resource_tailnet_key_test.go
+++ b/tailscale/resource_tailnet_key_test.go
@@ -67,6 +67,9 @@ func testTailnetKeyStruct(reusable bool) tailscale.Key {
 
 func setKeyStep(reusable bool, recreateIfInvalid string) resource.TestStep {
 	return resource.TestStep{
+		PreConfig: func() {
+			testServer.ResponseBody = testTailnetKeyStruct(reusable)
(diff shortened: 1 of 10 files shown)


tailscale/resource_tailnet_key_test.go
```diff
@@ -67,6 +67,9 @@ func testTailnetKeyStruct(reusable bool) tailscale.Key {
 
 func setKeyStep(reusable bool, recreateIfInvalid string) resource.TestStep {
 	return resource.TestStep{
+		PreConfig: func() {
+			testServer.ResponseBody = testTailnetKeyStruct(reusable)
```
mpminardi, line 71: This change just makes sure the server is ready to respond with the expected data immediately since we now do a `Read` after the `Create`.

willnorris (approved):

PR #363, opened 2024-05-03
fix imports for splitdns

keeleysam:
Currently if you use import on `tailscale_dns_split_nameservers`, the `domain` attribute will get set to `null` in the terraform state, and thus will result in 

Since the resource ID and domain are the same, set domain to the resource ID.
sample code:

```
import {
  to = tailscale_dns_split_nameservers.example-com
  id = "example.com"
}

resource "tailscale_dns_split_nameservers" "example-com" {
  domain      = "example.com"
  nameservers = [
        "1.2.3.4",
        "5.6.7.8"
  ]
}
```
Before:
```
# tailscale_dns_split_nameservers.example-com will be updated in-place
~ resource "tailscale_dns_split_nameservers" "example-com" {
    + domain      = "example.com"
      id          = "example.com"
      nameservers = [
        "1.2.3.4",
        "5.6.7.8"
      ]
  }
```

after:
```
  # tailscale_dns_split_nameservers.example-com will be imported
resource "tailscale_dns_split_nameservers" "example-com" {
    domain      = "example.com"
    id          = "example.com"
    nameservers = [
        "1.2.3.4",
        "5.6.7.8"
      ]
}
```

see above

### DIFF
--- a/tailscale/resource_dns_split_nameservers.go
+++ b/tailscale/resource_dns_split_nameservers.go
@@ -44,6 +44,12 @@ func resourceSplitDNSNameserversRead(ctx context.Context, d *schema.ResourceData
 		return diagnosticsError(err, "Failed to fetch split DNS configs")
 	}
 
+	domain := d.Id()
+
+	if err = d.Set("domain", domain); err != nil {
+		return diag.FromErr(err)
+	}
+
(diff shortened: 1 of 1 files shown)


mpminardi (approved):
Thank you for catching and fixing this!

PR #410, opened 2024-08-13
tailscale: removed unused data_source_devices

oxtoacart:
Updates #cleanup

This datasource was not referenced anywhere, so appears to be dead code.


oxtoacart:
Never mind, dataSourceDevices() is used.

PR #411, opened 2024-08-13
tailscale: add user and users data sources

oxtoacart:
These support accessing information about users via the User API.

Updates tailscale/corp#21629

Depends on https://github.com/tailscale/tailscale-client-go/pull/101


tailscale/provider.go
```diff
@@ -286,3 +288,14 @@ func readWithWaitFor(fn schema.ReadContextFunc) schema.ReadContextFunc {
 		}
 	}
 }
+
+// setProperties sets the properties of a ResourceData using the values in a
+// map.
+func setProperties(d *schema.ResourceData, props map[string]any) diag.Diagnostics {
```
soniaappasamy, line 294:
```suggestion
// setProperties sets the provided map props entries as properties of a ResourceData.
func setProperties(d *schema.ResourceData, props map[string]any) diag.Diagnostics {
```
take or leave, trying to make it a tad clearer that it doesn't clear out any existing properties, it's purely additive

tailscale/data_source_user.go
```diff
@@ -0,0 +1,118 @@
...
+	client := m.(*Clients).V2
+
+	id, hasID := d.Get("id").(string)
+	if !hasID {
+		return diagnosticsError(errors.New("data_source_user missing user ID"), "data_source_user missing user ID")
+	}
+
+	user, err := client.Users().Get(ctx, id)
+	if err != nil {
+		return diagnosticsError(err, "Failed to fetch user with id %s", id)
```
soniaappasamy, line 95:
```suggestion
		return diagnosticsError(err, "failed to fetch data_source_user with id %s", id)
```
not sure if we typically capitalize these errors

tailscale/data_source_users.go
```diff
@@ -0,0 +1,72 @@
...
+
+func dataSourceUsers() *schema.Resource {
+	return &schema.Resource{
+		Description: "The users data source describes a list of users in a tailnet",
+		ReadContext: dataSourceUsersRead,
+		Schema: map[string]*schema.Schema{
+			"type": {
+				Optional:    true,
+				Type:        schema.TypeString,
+				Description: "Filters the users list to elements whose type is the provided value.",
```
soniaappasamy, line 20: When these are enums, do we typically include the values anywhere in the description? Or do we leave that out to avoid getting out of date?

tailscale/data_source_users.go
```diff
@@ -0,0 +1,72 @@
...
+		},
+	}
+}
+
+func dataSourceUsersRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	client := m.(*Clients).V2
+
+	var userType *tsclient.UserType
+	_userType, hasType := d.Get("type").(string)
+	if hasType {
```
soniaappasamy, line 44: nit
```suggestion
	if _userType, ok := d.Get("type").(string); ok {
```

soniaappasamy: Could do same for `userRole` below

tailscale/data_source_user.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	d.SetId(user.ID)
+	return setProperties(d, UserToMap(user))
+}
+
+// UserToMap converts the given user into a map representing the user as a
+// resource in Terraform.
```
soniaappasamy, line 103: Add a comment that this omits the "id"?

tailscale/data_source_users_test.go
```diff
@@ -0,0 +1,115 @@
...
+							if v != expected {
+								return fmt.Errorf("wrong value of %s for user %s, want %q, got %q", prop, id, expected, v)
+							}
+						}
+					}
+
+					// Now set up user datasources
+					for id := range usersByID {
+						userDataSources.WriteString(fmt.Sprintf("\ndata \"tailscale_user\" \"%s\" {\n  id = \"%s\"\n}\n", id, id))
+					}
```
soniaappasamy, line 80: We do this for use in the next test I think? Maybe add a note about that

tailscale/data_source_users_test.go
```diff
@@ -0,0 +1,115 @@
...
+		ProviderFactories: testAccProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: userDataSources.String(),
+				Check: func(s *terraform.State) error {
+					client := testAccProvider.Meta().(*tailscale.Clients).V2
+					users, err := client.Users().List(context.Background(), nil, nil)
+					if err != nil {
+						return fmt.Errorf("unable to list users: %s", err)
+					}
```
soniaappasamy, line 99: I'm a little confused why we call the `List` endpoint again here instead of calling `client.Users().Get` with individual IDs and comparing values? Maybe I'm understanding what this test is for a little wrong.

soniaappasamy (approved):
Couple comments, but looks good!

tailscale/data_source_users.go
oxtoacart: Good catch, I had enums like [this](https://github.com/tailscale/terraform-provider-tailscale/pull/411) on here, but took them out because most of the properties are computed and don't allow validators. However, this is not a computed property.

tailscale/data_source_users_test.go
oxtoacart: Ha, it's probably confusing because the test is incomplete - I meant to compare the values of each property. Will fix.

oxtoacart:
Thanks for the good suggestions @soniaappasamy !

tailscale/data_source_user.go
```diff
@@ -0,0 +1,119 @@
...
+	"display_name": {
+		Type:        schema.TypeString,
+		Description: "The name of the user.",
+		Computed:    true,
+	},
+	"login_name": {
+		Type:        schema.TypeString,
+		Description: "The emailish login name of the user.",
+		Computed:    true,
+	},
```
mpminardi, line 29: I think it might make sense to allow for optionally specifying the `display_name` or `login_name` for specifying the user in addition to the id for better ergonomics / readability of the resulting Terraform code. This would be similar to  how we allow [name or hostname](https://github.com/tailscale/terraform-provider-tailscale/blob/main/tailscale/data_source_device.go#L20-L30) for finding a device.

On a related note: do we currently have the user ID surfaced somewhere in the UI? Or do people need to hit the API to get this information

oxtoacart: We do not surface the ID

PR #415, opened 2024-08-19
tailscale: add tailnet settings resource

oxtoacart:
Updates tailscale/corp#21628

Depends on https://github.com/tailscale/tailscale-client-go/pull/103


tailscale/resource_tailnet_settings.go
```diff
@@ -0,0 +1,140 @@
...
+	client := m.(*Clients).V2
+	if err := client.TailnetSettings().Update(ctx, settings); err != nil {
+		return diagnosticsError(err, "Failed to update tailnet settings")
+	}
+
+	return nil
+}
+
+func resourceTailnetSettingsDelete(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	// We don't know what the default values for Tailnet settings should be, so deleting is a noop.
```
soniaappasamy, line 138: The "we don't know" confuses me here - technically we do know the defaults we assign to new tailnets. Maybe it's less that we don't know and more that you can't really "delete" settings. (Reverting to defaults could be argued as being "deletion" but also could be argued as not)

tailscale/provider.go
```diff
@@ -300,3 +301,12 @@ func setProperties(d *schema.ResourceData, props map[string]any) diag.Diagnostic
 	}
 	return nil
 }
+
+// optional checks the given ResourceData for the value specified by key.
+// If the value is unchanged, it turns nil. Else, it returns a pointer to the value.
+func optional[T any](d *schema.ResourceData, key string) *T {
```
soniaappasamy, line 307: Nit
```suggestion
// optional returns a pointer to the value at key in the given resource only if the value has changed.
// If the value is unchanged, nil is returned.
func optional[T any](d *schema.ResourceData, key string) *T {
```

tailscale/resource_tailnet_settings.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	settings := tsclient.UpdateTailnetSettingsRequest{
+		DevicesApprovalOn:                      optional[bool](d, "devices_approval_on"),
+		DevicesAutoUpdatesOn:                   optional[bool](d, "devices_auto_updates_on"),
+		DevicesKeyDurationDays:                 optional[int](d, "devices_key_duration_days"),
+		UsersApprovalOn:                        optional[bool](d, "users_approval_on"),
+		UsersRoleAllowedToJoinExternalTailnets: role,
+		NetworkFlowLoggingOn:                   optional[bool](d, "network_flow_logging_on"),
+		RegionalRoutingOn:                      optional[bool](d, "regional_routing_on"),
+		PostureIdentityCollectionOn:            optional[bool](d, "posture_identity_collection_on"),
```
soniaappasamy, line 126: FWIW we could just update every field every time. Kinda just acting like a PUT. Fine either way.

soniaappasamy (approved):

tailscale/resource_tailnet_settings.go
oxtoacart: Maybe I should say "the terraform provider does not know the defaults". There's no API to "get defaults" or "reset to defaults", and I don't think we want to hardcode this in the terraform provider.

tailscale/resource_tailnet_settings.go
oxtoacart: We can't actually. Some fields require special plans (e.g. regional routing) and some require special permissions (e.g. network flow logging).

tailscale/resource_tailnet_settings.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+
+	tsclient "github.com/tailscale/tailscale-client-go/v2"
+)
+
+func resourceTailnetSettings() *schema.Resource {
+	return &schema.Resource{
+		Description:   "The tailnet_settings resource allows you to configure settings for your tailnet. See https://tailscale.com/api#tag/tailnetsettings/GET/tailnet/{tailnet}/settings for more information.",
```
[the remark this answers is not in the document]
oxtoacart: I did not, thanks!

tailscale/resource_tailnet_settings.go
```diff
@@ -0,0 +1,140 @@
...
+				Optional:    true,
+			},
+			"users_approval_on": {
+				Type:        schema.TypeBool,
+				Description: "Whether user approval is enabled for this tailnet",
+				Optional:    true,
+			},
+			"users_role_allowed_to_join_external_tailnet": {
+				Type:        schema.TypeString,
+				Description: "Which user roles are allowed to join external tailnets",
```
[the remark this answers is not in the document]
oxtoacart: This was taken verbatim from the API doc.

PR #419, opened 2024-08-20
tailscale: use V2 client for dns split nameservers resource

oxtoacart:
Updates tailscale/corp#21867

Depends on https://github.com/tailscale/tailscale-client-go/pull/104

### DIFF
--- a/tailscale/resource_dns_split_nameservers.go
+++ b/tailscale/resource_dns_split_nameservers.go
@@ -6,15 +6,15 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
-	"github.com/tailscale/tailscale-client-go/tailscale"
+	tsclient "github.com/tailscale/tailscale-client-go/v2"
 )
 
 func resourceDNSSplitNameservers() *schema.Resource {
 	return &schema.Resource{
 		Description:   "The dns_split_nameservers resource allows you to configure split DNS nameservers for your Tailscale network. See https://tailscale.com/kb/1054/dns for more information.",
 		ReadContext:   resourceSplitDNSNameserversRead,
-		CreateContext: resourceSplitDNSNameserversCreate,
-		UpdateContext: resourceSplitDNSNameserversUpdate,
+		CreateContext: resourceSplitDNSNameserversCreateOrUpdate,
+		UpdateContext: resourceSplitDNSNameserversCreateOrUpdate,
 		DeleteContext: resourceSplitDNSNameserversDelete,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
@@ -22,8 +22,9 @@ func resourceDNSSplitNameservers() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"domain": {
 				Type:        schema.TypeString,
-				Description: "Domain to configure split DNS for. Requests for this domain will be resolved using the provided nameservers.",
+				Description: "Domain to configure split DNS for. Requests for this domain will be resolved using the provided nameservers. Changing this will force the resource to be recreated.",
 				Required:    true,
+				ForceNew:    true,
 			},
 			"nameservers": {
 				Type:        schema.TypeSet,
@@ -38,8 +39,8 @@ func resourceDNSSplitNameservers() *schema.Resource {
 }
 
 func resourceSplitDNSNameserversRead(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-	client := m.(*Clients).V1
-	splitDNS, err := client.SplitDNS(ctx)
+	client := m.(*Clients).V2
+	splitDNS, err := client.DNS().SplitDNS(ctx)
 	if err != nil {
 		return diagnosticsError(err, "Failed to fetch split DNS configs")
 	}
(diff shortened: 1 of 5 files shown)


tailscale/resource_dns_split_nameservers.go
```diff
@@ -24,6 +24,7 @@ func resourceDNSSplitNameservers() *schema.Resource {
 				Type:        schema.TypeString,
 				Description: "Domain to configure split DNS for. Requests for this domain will be resolved using the provided nameservers.",
 				Required:    true,
+				ForceNew:    true,
```
oxtoacart, line 27: Drive-by: I noticed that this provider uses `UpdateSplitDNS()`, which is a patch operation. This means that the resource is not setting all split DNS settings for the tailnet, but just those for a specific domain. As originally implemented, if the domain changes, the settings for the previous domain will be left alone. That seems confusing.

Using ForceNew here ensures that the settings for the previous domain are removed, so that the state of the system accurately reflects what's configured in Terraform.

tailscale/resource_dns_split_nameservers.go
```diff
@@ -59,47 +60,39 @@ func resourceSplitDNSNameserversRead(ctx context.Context, d *schema.ResourceData
...
+	if _, err := client.DNS().UpdateSplitDNS(ctx, req); err != nil {
+		return diagnosticsError(err, "Failed to set create dns split nameservers")
 	}
 
 	d.SetId(domain)
 	return resourceSplitDNSNameserversRead(ctx, d, m)
 }
 
-func resourceSplitDNSNameserversUpdate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
-	if !d.HasChange("nameservers") {
```
oxtoacart, line 86: Because of [this](https://github.com/tailscale/terraform-provider-tailscale/pull/419/files#r1723858929), we no longer need to check for changes to nameservers.

tailscale/resource_dns_split_nameservers.go
```diff
@@ -6,24 +6,25 @@ import (
...
+		UpdateContext: resourceSplitDNSNameserversCreateOrUpdate,
 		DeleteContext: resourceSplitDNSNameserversDelete,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
 		Schema: map[string]*schema.Schema{
 			"domain": {
 				Type:        schema.TypeString,
-				Description: "Domain to configure split DNS for. Requests for this domain will be resolved using the provided nameservers.",
+				Description: "Domain to configure split DNS for. Requests for this domain will be resolved using the provided nameservers. This cannot be changed after the resource has been configured.",
```
tomhjp, line 25: nit: I can see what this update is trying to get at, but it makes it sound a bit like a permanent tailnet setting. I think the documentation is still ok without an update because users will get to see the "force new" behaviour in their plan output, but if you want to make it more explicit, how about something similar to "Updating this field will force the resource to be recreated"

tailscale/resource_dns_split_nameservers.go
tomhjp: Hmm I'm not deeply familiar with the underlying API but I don't quite follow; `"nameservers"` doesn't have `ForceNew: true`, so it seems like this field could be updated without recreating the resource?

oxtoacart: There are only two fields, `domain` and `nameservers`. Now that `domain` is `ForceNew`, any change to `domain` means that we'll create a new resource. If we're in `resourceSplitDNSNameserversCreateOrUpdate` and `domain` didn't change, that by definition means that `nameservers` must have changed, so there's no need to check for that explicitly.

tailscale/resource_dns_split_nameservers.go
oxtoacart: I'm going to wait for @mpminardi to chime in before merging, since this is a behavior change. From one perspective, it might be fixing buggy behavior, but possibly people could be relying on that buggy behavior ...

mpminardi: Whoops, good catch! Agree that this is buggy / unintended behaviour here that we should fix. Feel like this should be a safe change since we are essentially leaving a dangling resource outside of Terraform's control / lifecycle otherwise.

tailscale/resource_dns_split_nameservers.go
```diff
@@ -59,47 +60,39 @@ func resourceSplitDNSNameserversRead(ctx context.Context, d *schema.ResourceData
...
 	for _, nameserver := range nameserversList {
 		nameservers = append(nameservers, nameserver.(string))
 	}
 	req[domain] = nameservers
 
 	// Return value is not useful to us here, ignore.
-	if _, err := client.UpdateSplitDNS(ctx, req); err != nil {
-		return diagnosticsError(err, "Failed to set dns split nameservers")
+	if _, err := client.DNS().UpdateSplitDNS(ctx, req); err != nil {
+		return diagnosticsError(err, "Failed to set create dns split nameservers")
```
mpminardi, line 79: Was this error line intentionally changed?

mpminardi (approved):

tailscale/resource_dns_split_nameservers.go
oxtoacart: I've recently taken to differentiating between create and update flows. Of course, in this case, they're the same code. Will change back.

oxtoacart: Oh, the message had both create and set, oops :)

PR #421, opened 2024-08-23
tailscale: use V2 client for device and devices data sources

oxtoacart:
Updates tailscale/corp#21867

### DIFF
--- a/tailscale/data_source_device.go
+++ b/tailscale/data_source_device.go
@@ -9,7 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
-	"github.com/tailscale/tailscale-client-go/tailscale"
+	tsclient "github.com/tailscale/tailscale-client-go/v2"
 )
 
 func dataSourceDevice() *schema.Resource {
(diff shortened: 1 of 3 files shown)


tailscale/datasource_devices_test.go
```diff
@@ -0,0 +1,134 @@
...
+				Check: func(s *terraform.State) error {
+					client := testAccProvider.Meta().(*tailscale.Clients).V2
+					devices, err := client.Devices().List(context.Background())
+					if err != nil {
+						return fmt.Errorf("unable to list devices: %s", err)
+					}
+
+					devicesByID := make(map[string]map[string]any)
+					for _, device := range devices {
+						m := tailscale.DeviceToMap(&device)
```
tomhjp, line 42: nit: probably just personal opinion, but I'd prefer to copy-paste a small utility function into tests rather than make it exported purely for tests. Or just make this test part of the `tailscale` package rather than a dedicated `_test` package.

tailscale/data_source_device.go
```diff
@@ -108,26 +108,18 @@ func dataSourceDeviceRead(ctx context.Context, d *schema.ResourceData, m interfa
 	}
 
 	d.SetId(selected.ID)
+	return setProperties(d, DeviceToMap(selected))
```
tomhjp, line 111: Nice!

tailscale/datasource_devices_test.go
```diff
@@ -0,0 +1,134 @@
...
+						devicesByID[device.ID] = m
+					}
+
+					rs := s.RootModule().Resources[resourceName].Primary
+
+					// first find indexes for devices
+					deviceIndexes := make(map[string]string)
+					for k, v := range rs.Attributes {
+						if strings.HasSuffix(k, ".id") {
+							idx := strings.Split(k, ".")[1]
```
tomhjp, line 53: nit: For here and below, I always find it very helpful as a reader if detailed string manipulations are accompanied by an example or two in the comments of the expected structure that we're working on.

tailscale/datasource_devices_test.go
```diff
@@ -0,0 +1,134 @@
...
+								return fmt.Errorf("wrong value of %s for device %s, want %q, got %q", prop, id, expected, v)
+							}
+						}
+					}
+
+					// Now set up device datasources for each device. This is used in the following test
+					// of the tailscale_device datasource.
+					for _, device := range devices {
+						if device.Hostname != "" {
+							devicesDataSources.WriteString(fmt.Sprintf("\ndata \"tailscale_device\" \"%s\" {\n  hostname = \"%s\"\n}\n", toResourceComponent(device.Hostname), device.Hostname))
```
tomhjp, line 87: nit: This would be a lot easier to read with backtick strings, and you could probably also omit most of the `\n` characters too

```suggestion
							devicesDataSources.WriteString(fmt.Sprintf(`data "tailscale_device" "%s" {hostname = "%s"}\n`, toResourceComponent(device.Hostname), device.Hostname))
```

tomhjp (approved):
LGTM! Just a few nit suggestions you may or may not want to consider :)

tailscale/datasource_devices_test.go
oxtoacart: > Or just make this test part of the tailscale package rather than a dedicated _test package.

I started to go down that road, but there are reusable functions in the tailscale_test package that we want to use too. I'll look into maybe moving this into an `internal` package or something in a follow-on PR.

tailscale/datasource_devices_test.go
oxtoacart: For whatever reason, I actually prefer it on a single line to avoid weird indentation.

tomhjp: Yeah agreed about that. I more suggested backticks so the inner quotes don't need to be escaped with backslashes (less visual noise) but not a big deal either way

oxtoacart: Gotcha. AFAIK, with back ticks, I can't use `\n`, so I'd have to put it on multiple lines.

tomhjp: Ah of course, that makes sense now, thanks! 👍

PR #422, opened 2024-08-23
tailscale: use V2 client for acl data source

oxtoacart:
Updates tailscale/corp#21867

### DIFF
--- a/tailscale/data_source_acl_test.go
+++ b/tailscale/data_source_acl_test.go
@@ -1,0 +1,49 @@
+package tailscale_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+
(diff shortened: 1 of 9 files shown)


tailscale/data_source_acl_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+					huj, err := hujson.Parse([]byte(acl))
+					if err != nil {
+						return fmt.Errorf("Failed to parse ACL as HuJSON: %s", err)
+					}
+					expected := huj.String()
+
+					rs := s.RootModule().Resources[resourceName].Primary
+					actual := rs.Attributes["hujson"]
+					if diff := cmp.Diff(expected, actual); diff != "" {
```
tomhjp, line 40: nit: I think these might be the wrong way round? [ref](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff)

```suggestion
					if diff := cmp.Diff(actual, expected); diff != "" {
```

tomhjp (approved):

tailscale/data_source_acl_test.go
oxtoacart: Ha, looks like we have a lot of these. I miss using `stretchr/testify` 😭

PR #423, opened 2024-08-23
tailscale: rename tailscale_test package to tailscale

oxtoacart:
This allows the reuse of private functions like deviceToMap between the core package and the tests.

Updates tailscale/corp#21867


tomhjp (approved):
Makes sense to me 👍

mpminardi (approved):

PR #427, opened 2024-09-05
tailscale: allow configuring user data source using login_name or dis…

oxtoacart:
…play_name

This is useful because clients typically have no knowledge of user IDs.

Updates tailscale/corp#21867


tailscale/data_source_user.go
```diff
@@ -78,32 +68,78 @@ func dataSourceUser() *schema.Resource {
 	return &schema.Resource{
 		Description: "The user data source describes a single user in a tailnet",
 		ReadContext: dataSourceUserRead,
-		Schema:      userSchema,
+		Schema: combinedSchemas(commonUserSchema, map[string]*schema.Schema{
+			"login_name": {
+				Type:         schema.TypeString,
+				Description:  "The emailish login name of the user.",
+				Optional:     true,
+				AtLeastOneOf: []string{"display_name", "login_name"},
```
mpminardi, line 76: Should this be `ExactlyOneOf` instead of `AtLeastOneOf`, or are their cases where a user would want to be specifying both and only specifying one wouldn't be enough to identify a single user?

oxtoacart: Per discussion, I'm only allowing use of `login_name` now.

tailscale/data_source_user.go
```diff
@@ -78,32 +72,65 @@ func dataSourceUser() *schema.Resource {
...
 		Description: "The user data source describes a single user in a tailnet",
 		ReadContext: dataSourceUserRead,
-		Schema:      userSchema,
+		Schema: combinedSchemas(commonUserSchema, map[string]*schema.Schema{
+			"login_name": {
+				Type:        schema.TypeString,
+				Description: "The emailish login name of the user.",
+				Optional:    true,
+			},
+			"display_name": {
```
mpminardi, line 81: Think we'll want to remove this now that it is back in the `commonUserSchema`

tailscale/data_source_user.go
```diff
@@ -78,32 +72,65 @@ func dataSourceUser() *schema.Resource {
 	return &schema.Resource{
 		Description: "The user data source describes a single user in a tailnet",
 		ReadContext: dataSourceUserRead,
-		Schema:      userSchema,
+		Schema: combinedSchemas(commonUserSchema, map[string]*schema.Schema{
+			"login_name": {
+				Type:        schema.TypeString,
+				Description: "The emailish login name of the user.",
+				Optional:    true,
```
mpminardi, line 79: Now that this is the only way to query a user:
```suggestion
				Required:    true,
```

oxtoacart: It occurred to me that I'm still allowing use of ID, so I've updated the schema to make this clear.

mpminardi (approved):

PR #442, opened 2024-10-04
Add missing importer logic

### LINKED ISSUE #441: Support import for all tailscale terraform resources
I'm managing my personal tailscale configuration with terraform.
For ease of use it'd be preferable to support imports for the resources I am using, so I can get started configuring on the UI, then bring over the config to terraform and keep managing the resources. For that import would be preferable.
Without imports there is no clear diff between the existing config and the desired one made in terraform.

The resources I want to be able to import but lack import functionality:

Can be imported with any random id (only a single resource exists for it in the tailscale api):
```
tailscale_dns_nameservers
tailscale_dns_search_paths
```

Can be imported with an existing device id:

```
tailscale_device_key
tailscale_device_subnet_routes
tailscale_device_tags
```

IMHO allowing straight pass through id for the import is sufficient for all of the above resources.

**Additional context**
There are likely some other resources that also lack import support right now.

reegnz:

Some tailscale terraform resources were missing import capabilities.

**Which issue this PR fixes** _(use `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)_:

Fixes #441

### DIFF
--- a/docs/resources/device_key.md
+++ b/docs/resources/device_key.md
@@ -37,3 +37,12 @@ resource "tailscale_device_key" "example_key" {
 ### Read-Only
 
 - `id` (String) The ID of this resource.
+
+## Import
+
+Import is supported using the following syntax:
+
+```shell
+# Device key can be imported using the device id, e.g.,
+terraform import tailscale_device_key.sample 123456789
+```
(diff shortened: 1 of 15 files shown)


reegnz:
I have tested the import functionality for each resource with a locally built provider.

mpminardi (approved):
Thank you for adding these! Much appreciated.

PR #468, opened 2025-01-22
tailscale: add support for S3 logstreaming

### LINKED ISSUE #458: Support S3 log streaming

S3 log streaming is now an option. We should add the JSON body

zehauser:
We recently added support for S3 logstreaming endpoints to our API. This involved adding several new fields to the LogstreamConfiguration resource (available when the new "s3" destinationType is used), plus a new AWSExternalID resource needed when AWS role-based authentication is used for an S3 logstreaming endpoint.

This commit updates the Terraform provider to reflect these changes. We add support for the new fields to the tailscale_logstream_configuration resource, and create a new tailscale_aws_external_id resource.

Unfortunately, we have to add some special handling for the "user" field of tailscale_logstream_configuration. Previously, we specified a default dummy value of "user" for the "user" field. Now, the PUT LogstreamConfiguration endpoint does not allow any value to be specified for a s3 destinationType. We could loosen that restriction at the API layer, but I would prefer not to do that. However, we don't want to break any existing users who are relying on this default and have it in their Terraform state. So the approach I've chosen is to add special handling for the "user" field:
- When creating or updating a configuration, if destinationType == "s3" and user == "user", we assume the "user" value came from the default and we do not actually send it to the API.
- When reading or importing a configuration, if destinationType == "s3" and there is no user field (which should always be true), we set user = "user" in the Terraform state.

Fixes #458
Fixes tailscale/corp#24533


go.sum
```diff
@@ -188,8 +188,8 @@ github.com/stretchr/testify v1.9.0 h1:HtqpIVDClZ4nwg75+f6Lvsy/wHu+3BoSGCbBAcpTsT
 github.com/stretchr/testify v1.9.0/go.mod h1:r2ic/lqez/lEtzL7wO/rwa5dbSLXVDPFyf8C91i36aY=
 github.com/tailscale/hujson v0.0.0-20221223112325-20486734a56a h1:SJy1Pu0eH1C29XwJucQo73FrleVK6t4kYz4NVhp34Yw=
 github.com/tailscale/hujson v0.0.0-20221223112325-20486734a56a/go.mod h1:DFSS3NAGHthKo1gTlmEcSBiZrRJXi28rLNd/1udP1c8=
-github.com/tailscale/tailscale-client-go/v2 v2.0.0-20241028210109-bd4d815eb293 h1:qguJy85YoqTEdGznTqbOSh+GZgTvBpOCriN9/HhayK0=
-github.com/tailscale/tailscale-client-go/v2 v2.0.0-20241028210109-bd4d815eb293/go.mod h1:i/MSgQ71kdyh1Wdp50XxrIgtsyO4uZ2SZSPd83lGKHM=
+github.com/tailscale/tailscale-client-go/v2 v2.0.0-20250122081626-c84613baf18d h1:I3Ka+YodSpslWRaHHOtKQiHFcfuTr8o23619oP1P11I=
+github.com/tailscale/tailscale-client-go/v2 v2.0.0-20250122081626-c84613baf18d/go.mod h1:i/MSgQ71kdyh1Wdp50XxrIgtsyO4uZ2SZSPd83lGKHM=
```
zehauser, line 192: We probably need to modify this after https://github.com/tailscale/tailscale-client-go/pull/131 is actually merged.

tailscale/resource_logstream_configuration.go
```diff
@@ -101,6 +176,11 @@ func resourceLogstreamConfigurationRead(ctx context.Context, d *schema.ResourceD
 	client := m.(*tsclient.Client)
 
 	logstream, err := client.Logging().LogstreamConfiguration(ctx, tsclient.LogType(d.Id()))
+	if err != nil && tsclient.IsNotFound(err) {
+		d.SetId("")
+		return nil
+	}
```
zehauser, line 182: Oh right! This is an unrelated bugfix I snuck in. Or at least I think it's a bugfix.

If you delete your Terraform-created logstream configuration out-of-band (e.g. in the admin console), without this fix, we fail with a 404 error on `plan`. I think probably we should instead detect that the config has been deleted and report that, right?

zehauser (commented):
Successful acceptance test run: https://github.com/tailscale/corp/actions/runs/12906014118

tailscale/resource_aws_external_id.go
```diff
@@ -0,0 +1,56 @@
...
+				Description: "The AWS account from which Tailscale will assume your role. You probably want to reference this in your IAM role's trust policy.",
+			},
+		},
+	}
+}
+
+func resourceAWSExternalIDCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	client := m.(*tsclient.Client)
+
+	aid, err := client.Logging().CreateOrGetAwsExternalId(ctx, tsclient.CreateOrGetAwsExternalIdRequest{Reusable: false})
```
oxtoacart, line 42: It might be worth a comment on why "reusable" is false, and why it must be.

tailscale/resource_logstream_configuration.go
```diff
@@ -77,16 +126,42 @@ func resourceLogstreamConfigurationCreate(ctx context.Context, d *schema.Resourc
...
+	s3ExternalID := d.Get("s3_external_id").(string)
+
+	// We'll get an error if we send the user field to the API with destination_type = 's3'.
+	// But unfortunately, prior to introducing S3 logstreaming, we specified a default value
+	// of user = 'user' in the Terraform provider, and we don't want to break existing Terraform
+	// state. So we'll avoid sending the user value to the API in this scenario. (And in the
+	// corresponding scenario for resourceLogstreamConfigurationRead(), we'll add
+	// user = 'user' to the state.)
+	user := d.Get("user").(string)
+	if destinationType == tsclient.LogstreamS3Endpoint && user == "user" {
```
oxtoacart, line 148: Do we need the `user == "user"` check? For S3, we never want to send a user, right?

tailscale/resource_logstream_configuration.go
oxtoacart: I'll defer to @mpminardi on this, as I don't know what the usual convention is with Terraform.

tailscale/resource_logstream_configuration.go
oxtoacart: And we if change this, can we get rid of [the below](https://github.com/tailscale/terraform-provider-tailscale/pull/468/files#diff-69a18e175ae577b16431d6c966bc7360a1295106a5da4a49eda6661834b61eefR207)?

tailscale/resource_logstream_configuration_test.go
```diff
@@ -38,7 +39,33 @@ const testLogstreamConfigurationUpdateDifferentLogtype = `
...
+	}
+	resource "tailscale_aws_external_id" "external_id" {}
+	`
+
+const testLogstreamConfigurationUpdateS3AccessKey = `
+		resource "tailscale_logstream_configuration" "test_logstream_configuration" {
+			log_type               = "network"
+			destination_type       = "s3"
+			s3_bucket              = "example-bucket"
+			s3_region			   = "us-west-2"
```
oxtoacart, line 61: Nit: weird indentation

oxtoacart (commented):
Finished my first round of review.

tailscale/resource_logstream_configuration.go
mpminardi: Whoops good catch on this! Yes I think if we get a 404 here we do want to report this / remove the resource from state. Have filed [this](https://github.com/tailscale/corp/issues/26072) ticket to track doing this across the board as I imagine we are missing this in a number of places.

tailscale/resource_aws_external_id.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	tsclient "github.com/tailscale/tailscale-client-go/v2"
+)
+
+func resourceAWSExternalID() *schema.Resource {
+	return &schema.Resource{
+		Description:   "The aws_external_id resource allows you to mint an AWS External ID that Tailscale can use to assume an AWS IAM role that you create for the purposes of allowing Tailscale to stream logs to your S3 bucket. See the logstream_configuration resource for more.",
```
mpminardi, line 17:
```suggestion
		Description:   "The aws_external_id resource allows you to mint an AWS External ID that Tailscale can use to assume an AWS IAM role that you create for the purposes of allowing Tailscale to stream logs to your S3 bucket. See the logstream_configuration resource for more details.",
```

tailscale/resource_aws_external_id.go
```diff
@@ -0,0 +1,56 @@
...
+
+		// No GET or DELETE endpoints in the API. This is a create-only resource.
+		ReadContext:   schema.NoopContext,
+		DeleteContext: schema.NoopContext,
+
+		Schema: map[string]*schema.Schema{
+			"external_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The actual External ID that Tailscale will supply when assuming your role. You probably want to reference this in your IAM role's trust policy.",
```
mpminardi, line 28: I think we will want to be more prescriptive / strong in our language here and potentially reference the AWS documentation / guidance for external IDs:

```suggestion
				Description: "The External ID that Tailscale will supply when assuming your role. You must reference this in your IAM role's trust policy. See https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_common-scenarios_third-party.html for more information on external IDs.",
```

tailscale/resource_aws_external_id.go
```diff
@@ -0,0 +1,56 @@
...
+			},
+			"tailscale_aws_account_id": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The AWS account from which Tailscale will assume your role. You probably want to reference this in your IAM role's trust policy.",
```
mpminardi, line 33: Same here as above, I think we want stronger wording:
```suggestion
				Description: "The AWS account from which Tailscale will assume your role. You must reference this in your IAM role's trust policy. See https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_common-scenarios_third-party.html for more information on external IDs.",
```

tailscale/resource_logstream_configuration.go
zehauser: With the current approach, I think we do need the `user == "user"` check both here and in `resourceLogstreamConfigurationRead()`. With that check, if a user explicitly specifies something else...
```
resource tailscale_logstream_configuration "config" {
    destination_type = "s3"
    user = "foobar"
}
```
...then they'll get an API error at apply time saying that `user` cannot be provided when `destinationType` is `"s3"`. Which is fine.

If we drop the `user == "user"` check, then the user won't get an error at apply time because we'll silently omit the `user` field from the PUT request. But then something more insidious will happen — at *plan* time, they'll get a spurious diff, because `resourceLogstreamConfigurationRead()` will report `user = "user"` (or no `user`, depending on the specific logic), while the resource in the state has `user = "foobar"`. We can't report `user = "foobar"` from `resourceLogstreamConfigurationRead()` because the info that the user specified `"foobar"` isn't available in that context.

zehauser: Let me know if you think I'm totally confused about this and should change it!

tailscale/resource_aws_external_id.go
zehauser: @oxtoacart I pushed up [a change](https://github.com/tailscale/terraform-provider-tailscale/compare/21d8c90f1a7dc5bb0a3392391ef29f518266c72d..6415cdae90931e7b5febd0ef2ca627362da32ba2#diff-4323850be6e4e1e73caf830d8915901a6ba80d9a2318c892af3fe5b2e0f29e2cR42) that includes a (perhaps overlong) comment about this. Take a look and let me know if you think it's helpful.

But it's possible that I'm missing the point here, because I actually don't see why it _must_ be: the reasons I see are that there's no real need, and in certain circumstances (not that common, though) it could lead to errors. 

If I'm missing the point and there's a much simpler explanation please let me know.

oxtoacart: I think(?) the explanation is something like this:

We have to set `reusable` to `false`. Otherwise, multiple external ID resources could end up with the same Terraform ID.

Side note - I don't fully understand the implications of two different resources having the same ID, but I suspect it's not good.

tailscale/resource_logstream_configuration.go
oxtoacart: That makes sense, thank you!

oxtoacart (approved):
Just one outstanding question on the comment related to external ID reuse, but not a blocker.

tailscale/resource_aws_external_id.go
zehauser: Updated!

zehauser: And thanks for explaining. I think you're right about that. Resource IDs are supposed to be unique and Terraform would be unhappy.

go.sum
zehauser: Done. Although #477 is about to make this moot.

zehauser:
I rebased on `main`, updated to the latest https://github.com/tailscale/tailscale-client-go, and updated a comment per Percy's helpful explanation.

More significantly, due to our recent API consistency fix, I was able to drop all the special handling around the `user = "user"` default.

I'll merge after running acceptance tests.

zehauser:
Acceptance tests: https://github.com/tailscale/corp/actions/runs/13193331254/job/36830140348

I also mildly improved (IMO) the example Terraform configs.

And finally, I tested a full RoleARN-authentication configuration with my personal tailnet — quite similar to the example config in the `tailscale_aws_external_id` docs — where I created IAM role + policies and Tailscale logging config + external ID from scratch via Terraform, confirmed that streaming worked, and then tested `terraform destroy`.

PR #500, opened 2025-04-29
tailscale: add oauth client resource

mcoulombe:
Added support to manage OAuth clients via the tailscale_oauth_client resource 
Updates: tailscale/tailscale#9632

Acc test results on the branch:
https://github.com/tailscale/corp/actions/runs/14762878480


tailscale/resource_tailnet_key.go
```diff
@@ -128,6 +128,7 @@ func resourceTailnetKeyCreate(ctx context.Context, d *schema.ResourceData, m int
 	}
 
 	var req tailscale.CreateKeyRequest
+	req.Capabilities = &tailscale.KeyCapabilities{}
```
mcoulombe, line 131: Consequence of making the struct a pointer to make it optional on the API.

tailscale/resource_oauth_client.go
```diff
@@ -0,0 +1,155 @@
...
+	"time"
+
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"tailscale.com/client/tailscale/v2"
+)
+
+func resourceOAuthClient() *schema.Resource {
```
oxtoacart, line 17: I'm thinking through the consequences of making this a separate resource from `tailnet_key`.

`tailnet_key` can be [imported](https://registry.terraform.io/providers/tailscale/tailscale/latest/docs/resources/tailnet_key#import) using just a key ID. In theory, `tailnet_key` would allow importing what's actually an OAuth client. Could we run into some weird state inconsistencies if someone does this?

@mpminardi thoughts?

oxtoacart: Maybe we should update the `tailnet_key` resource to fail if someone tries to import something that is not an auth key?

mcoulombe: I don't think it can lead to TF state inconsistencies, or well, no inconsistencies worse than what Terraform naturally allows.

Terraform lets you to import the same external resource multiple times for different resource definitions. E.g.:
```
tf import tailscale_tailnet_key.import1 k1234DEVEL
tf import tailscale_tailnet_key.import2 k1234DEVEL
```
Is considered perfectly valid and will create 2 state entries for the same resource. So whether tailscale_tailnet_key is extended to support oauth clients or we implement it as a separate resource does not really matter wrt how imports can be misused.

Fwiw imo importing `tailscale_tailnet_key` should not have been allowed at all. Since keys cannot be modified and importing does not return the secret key, I don't see any good import use case. Users can even import access tokens since they are technically also keys 😅

oxtoacart (commented):
Generally LGTM, just a few suggestions on naming. Would also be nice to have @mpminardi weigh in.

tailscale/resource_oauth_client.go
```diff
@@ -0,0 +1,154 @@
...
+			"tags": {
+				Type: schema.TypeSet,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Optional:    true,
+				Description: "A list of tags that access tokens generated for the OAuth client will be able to assign to devices. Mandatory is the scopes include \"devices:core\" or \"auth_keys\".",
+				ForceNew:    true,
+			},
+			"id": {
```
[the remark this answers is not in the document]
mpminardi: In this case the `client_id` is the same as / being used as the actual ID of the resource. I think I would lean more towards leaving this and the `key` attributes named as they are to stay more consistent with the structure of the underlying API.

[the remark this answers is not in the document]
mcoulombe: I went for parity with the API (id/key pair). I do agree though that if we had a magic wand or V3 API we should align the field names with the oauth spec.

tailscale/resource_oauth_client.go
```diff
@@ -0,0 +1,154 @@
...
+				Type:        schema.TypeString,
+				Description: "The client ID, also known as the key id. Used with the client secret to generate access tokens.",
+				Computed:    true,
+				Sensitive:   true,
```
mpminardi, line 60: Don't think this needs to be sensitive! We return this in the API / show it in the UI

tailscale/resource_oauth_client.go
```diff
@@ -0,0 +1,154 @@
...
+			},
+			"created_at": {
+				Type:        schema.TypeString,
+				Description: "The creation timestamp of the key in RFC3339 format",
+				Computed:    true,
+			},
+			"user_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "ID of the user who created this key, or empty if not created by a user.",
```
mpminardi, line 76: Think it might be helpful to add some colour around when this is the case:

```suggestion
				Description: "ID of the user who created this key, empty for OAuth clients created by other OAuth clients.",
```

tailscale/resource_oauth_client.go
```diff
@@ -0,0 +1,154 @@
...
+		return diagnosticsError(err, "Failed to set key")
+	}
+	if err = d.Set("created_at", key.Created.Format(time.RFC3339)); err != nil {
+		return diagnosticsError(err, "Failed to set created_at")
+	}
+	if err = d.Set("user_id", key.UserID); err != nil {
+		return diagnosticsError(err, "Failed to set user_id")
+	}
+
+	return nil
```
mpminardi, line 142: As a sanity measure we usually do a read after create to protect against potential [state drift issues](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/detecting-drift#update-state-after-modification):

```suggestion
	return resourceOAuthClientRead(ctx, d, m)
```

mpminardi (approved):

PR #503, opened 2025-05-06
tailscale: prevent import of non-auth keys

### LINKED ISSUE #502: tailscale_tailnet_key shouldn't allow importing non-auth keys
**Describe the bug**
It should not be possible to import non-auth keys for a tailscale_tailnet_key resource.

**To Reproduce**
1. Manually create a OAuth client or API access key in the Tailscale UI
2. Define a `tailscale_tailnet_key` resource
3. terraform import tailscale_tailnet_key.my_key <key-id-from-a-non-auth-key>
4. Your state now manages a resource that cannot be described by the resource, leading to unexpected behaviors

**Expected behaviour**
The provider should error when trying to import a non-auth key.

mcoulombe:
Importing a non-auth key creates an inconsistent statefile. The resource ID references a key that doesn't match the tailscale_tailnet_key schema so it is impossible to satisfy the diff. The apply will _always_ delete the imported key and create a new one so allowing imports serves no purpose and is a potential footgun.

Fixes #502

Do not merge before this is closed and the provider pulls on the proper client commit:
https://github.com/tailscale/tailscale-client-go-v2/pull/26

Example:
Screenshot 2025-05-06 at 9 44 44 AM


go.mod
```diff
@@ -14,9 +14,11 @@ require (
 	github.com/tailscale/hujson v0.0.0-20221223112325-20486734a56a
 	golang.org/x/tools v0.32.0
 	tailscale.com v1.80.2
-	tailscale.com/client/tailscale/v2 v2.0.0-20250502205821-61a211e0f308
+	tailscale.com/client/tailscale/v2 v2.0.0-20250506130935-55568c07697b // TODO use the ral main commit before merging
```
mcoulombe, line 17: https://github.com/tailscale/tailscale-client-go-v2/pull/26

tailscale/resource_oauth_client.go
```diff
@@ -20,7 +20,7 @@ func resourceOAuthClient() *schema.Resource {
 		ReadContext:   resourceOAuthClientRead,
 		CreateContext: resourceOAuthClientCreate,
 		DeleteContext: resourceOAuthClientDelete,
-		UpdateContext: schema.NoopContext,
+		UpdateContext: nil,
```
mcoulombe, line 23: Since the user_id is not optional anymore we have to unset the UpdateContext as every fields is either ForceNew or Computed. Otherwise Terraform complains with:
`All fields are ForceNew or Computed without Optional, Update is superfluous`

tailscale/resource_oauth_client.go
```diff
@@ -71,7 +71,6 @@ func resourceOAuthClient() *schema.Resource {
 			},
 			"user_id": {
 				Type:        schema.TypeString,
-				Optional:    true,
```
mcoulombe, line 74: Marking the field as optional incorrectly documented is as user-configurable whereas we want it to be a documented as a read-only field.

mpminardi (approved):

PR #513, opened 2025-05-30
tailscale: add node_id field to device data sources

### LINKED ISSUE #511: FR: Add support for NodeID in `data_source_device` and `data_source_devices`
### **Is your feature request related to a problem? Please describe.**
Currently the device data sources do not contain the [nodeID](https://github.com/tailscale/tailscale-client-go-v2/blob/5fad10cf3a3372233783436c03e6b77d3d0791b2/devices.go#L74) which is the preferred identifier to use. The only thing exposed today is [ID](https://github.com/tailscale/tailscale-client-go-v2/blob/5fad10cf3a3372233783436c03e6b77d3d0791b2/devices.go#L73) which appears to be the legacy ID no longer recommended for use.

### **Describe the solution you'd like**
I would like to be able to reference the nodeID/have it exposed for these data source resources.

zbuchheit:
Adds `node_id` to both `tailscale_device` and `tailscale_devices` data sources. Currently only the legacy id is exposed to users.

Fixes #511


tailscale/data_source_devices.go
```diff
@@ -46,7 +46,12 @@ func dataSourceDevices() *schema.Resource {
 						},
 						"id": {
 							Type:        schema.TypeString,
-							Description: "The unique identifier of the device",
+							Description: "The legacy identifier of the device. Use node_id instead.",
```
mpminardi: Think we'll want to caveat this for specifically new resources:
```suggestion
							Description: "The legacy identifier of the device. Use node_id instead for new resources.",
```

PR #547, opened 2025-09-17
resource_dns_configuration: create resource for managing entire DNS c…

oxtoacart:
…onfiguration at once

This resource uses the new `/tailnet/-/dns/configuration` endpoint to manage the complete DNS configuration for a tailnet.

Updates tailscale/corp#32130

Depends on https://github.com/tailscale/tailscale-client-go-v2/pull/47


tailscale/resource_dns_configuration.go
```diff
@@ -0,0 +1,257 @@
...
+				Optional:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"address": {
+							Description: "The nameserver's IPv4 or IPv6 address",
+							Type:        schema.TypeString,
+							Required:    true,
+						},
+						"use_with_exit_node": {
+							Description: "This name server will continue to be used when an exit node is selected (requires Tailscale v1.88 or later).",
```
mpminardi: Nit:
```suggestion
							Description: "This name server will continue to be used when an exit node is selected (requires Tailscale v1.88.1 or later).",
```

tailscale/resource_dns_configuration.go
```diff
@@ -0,0 +1,257 @@
...
+							Required:    true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"address": {
+										Description: "The nameserver's IPv4 or IPv6 address.",
+										Type:        schema.TypeString,
+										Required:    true,
+									},
+									"use_with_exit_node": {
+										Description: "This name server will continue to be used when an exit node is selected (requires Tailscale v1.88 or later).",
```
mpminardi:
```suggestion
										Description: "This name server will continue to be used when an exit node is selected (requires Tailscale v1.88.1 or later).",
```

tailscale/resource_dns_configuration.go
```diff
@@ -0,0 +1,257 @@
...
+				},
+				Optional: true,
+			},
+			"override_local_dns": {
+				Description: "When enabled, use the configured DNS servers to resolve names outside the tailnet. When disabled, devices will prefer their local DNS configuration.",
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+			},
+			"disable_magic_dns": {
```
mpminardi: Is there a reason we have this flipped from the API?

tailscale/resource_dns_configuration.go
```diff
@@ -0,0 +1,257 @@
...
+							Description: "This name server will continue to be used when an exit node is selected (requires Tailscale v1.88 or later).",
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Default:     false,
+						},
+					},
+				},
+			},
+			"split_dns": {
+				Description: "Set the nameservers used by devices on your network to resolve DNS queries on specific domains (requires Tailscale v1.8 or later).",
```
mpminardi: Was the version disclaimer here intentional?

tailscale/resource_dns_configuration.go
oxtoacart: Good question. I was just going by what was in the design. @mzbenami thoughts?

go.sum
```diff
@@ -312,3 +312,9 @@ tailscale.com v1.86.4 h1:KHAmLyzVn50t2P5877wohBU0UPVeIMHC9XDzRw4Ycz4=
 tailscale.com v1.86.4/go.mod h1:Lm8dnzU2i/Emw15r6sl3FRNp/liSQ/nYw6ZSQvIdZ1M=
 tailscale.com/client/tailscale/v2 v2.0.0-20250616133344-8dcb33eb281b h1:Cyso/184f0BUfEDUWVmM65t6byUEB8wuTBGpB8Tv1PA=
 tailscale.com/client/tailscale/v2 v2.0.0-20250616133344-8dcb33eb281b/go.mod h1:nzqx3Hs59z2W8Gnmq2ChavPButcyvtxAxRpNc+ZVy7s=
+tailscale.com/client/tailscale/v2 v2.0.0-20250916195646-d12591c12412 h1:NgzOCjJxjEWKKWiL3C92wGvml/c19OEe56g+U+h3L1w=
+tailscale.com/client/tailscale/v2 v2.0.0-20250916195646-d12591c12412/go.mod h1:4akEJPbysqHWAP+t7CZLQ5ZH8/vZWeH6+Hv+fEJUMp0=
+tailscale.com/client/tailscale/v2 v2.0.0-20250917004044-bfc0c6287afa h1:lwF031L0zMDaLW67+5xE6z493ehJszuh935CXX3RON4=
+tailscale.com/client/tailscale/v2 v2.0.0-20250917004044-bfc0c6287afa/go.mod h1:4akEJPbysqHWAP+t7CZLQ5ZH8/vZWeH6+Hv+fEJUMp0=
+tailscale.com/client/tailscale/v2 v2.0.0-20250917004339-7860180eaebc h1:PEtgrcOJ8GSQh2Eg8AXqLPDof1gGx9Dr3cAxZjp/rrM=
+tailscale.com/client/tailscale/v2 v2.0.0-20250917004339-7860180eaebc/go.mod h1:4akEJPbysqHWAP+t7CZLQ5ZH8/vZWeH6+Hv+fEJUMp0=
```
mpminardi: Reminder to `go mod tidy` once the client PR is through

tailscale/resource_dns_configuration.go
oxtoacart: Taken from UI

image

tailscale/resource_dns_configuration.go
mzbenami: Sorry I didn't do a great job circling back to the doc after I made some decisions, but I think the two open PRs for the API and Go client can serve as the source of truth over the doc, unless it violates some TF-specific best practices.

tailscale/resource_dns_configuration.go
oxtoacart: The UI says 1.88, is it wrong?

image

mpminardi: `v1.88.1` was the publicly tagged / released version of `v1.88`. Think we can leave this as-is if we want to be less specific (and in case another patch release comes out).

docs/resources/dns_configuration.md
```diff
@@ -0,0 +1,59 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
```
mpminardi: We can tweak the template that is used to generate this to get some nicer formatted output / a proper "warning" block rendered by the Terraform registry. I think if we stick something that looks like:

```
---
page_title: "{{.Name}} {{.Type}} - {{.ProviderName}}"
subcategory: ""
description: |-
{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
---

# {{.Name}} ({{.Type}})

{{ .Description | trimspace }}

~> **Warning:** This resource is in alpha, proceed with caution!

~> **Note:** The Tailscale Terraform provider has multiple resources for managing DNS configuration. This resource  is meant to manage the entirety of a Tailnet's DNS configuration and conflicts with [tailscale_dns_nameservers](dns_nameservers.md), [tailscale_dns_preferences](dns_preferences.md), [tailscale_dns_search_paths](dns_search_paths.md), and [tailscale_dns_split_nameservers](dns_split_nameservers.md). This resource and previously mentioned resources cannot be used simultaneously.

## Example Usage

{{ tffile (printf "examples/resources/%s/resource.tf" .Name)}}

{{ .SchemaMarkdown | trimspace }}

## Import

Import is supported using the following syntax:

{{ codefile "shell" (printf "examples/resources/%s/import.sh" .Name)}}

```

into `templates/resources/dns_configuration.md.tmpl` that should do the trick.

mpminardi: We'll also want to add the `resource.tf` and `import.sh` examples here.

oxtoacart: Done. I changed "cannot be used" to "should not be used", as we don't enforce this, right?

tailscale/resource_dns_configuration.go
oxtoacart: I'll leave it as-is for consistency with the UI.

docs/resources/dns_configuration.md
mpminardi: Yeah we don't have a mechanism to enforce this, but "cannot" in the sense that they can't be used at the same time without there being repeated state drift between the competing resources

tailscale/resource_dns_configuration.go
```diff
@@ -0,0 +1,257 @@
...
+				Optional:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"address": {
+							Description: "The nameserver's IPv4 or IPv6 address",
+							Type:        schema.TypeString,
+							Required:    true,
+						},
+						"use_with_exit_node": {
+							Description: "This name server will continue to be used when an exit node is selected (requires Tailscale v1.88 or later).",
```
mpminardi:
```suggestion
							Description: "This name server will continue to be used when an exit node is selected (requires Tailscale v1.88 or later). Defaults to `false`.",
```

tailscale/resource_dns_configuration.go
```diff
@@ -0,0 +1,257 @@
...
+							Required:    true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"address": {
+										Description: "The nameserver's IPv4 or IPv6 address.",
+										Type:        schema.TypeString,
+										Required:    true,
+									},
+									"use_with_exit_node": {
+										Description: "This name server will continue to be used when an exit node is selected (requires Tailscale v1.88 or later).",
```
mpminardi:
```suggestion
										Description: "This name server will continue to be used when an exit node is selected (requires Tailscale v1.88 or later). Defaults to `false`.",
```

tailscale/resource_dns_configuration.go
```diff
@@ -0,0 +1,257 @@
...
+			"search_paths": {
+				Description: "Additional search domains. When MagicDNS is on, the tailnet domain is automatically included as the first search domain.",
+				Type:        schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Optional: true,
+			},
+			"override_local_dns": {
+				Description: "When enabled, use the configured DNS servers to resolve names outside the tailnet. When disabled, devices will prefer their local DNS configuration.",
```
mpminardi:
```suggestion
				Description: "When enabled, use the configured DNS servers to resolve names outside the tailnet. When disabled, devices will prefer their local DNS configuration. Defaults to `false`.",
```

tailscale/resource_dns_configuration.go
```diff
@@ -0,0 +1,257 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+			},
+			"magic_dns": {
+				Description: "Whether or not to enable magic DNS.",
```
mpminardi:
```suggestion
				Description: "Whether or not to enable magic DNS. Defaults to `true`.",
```

tailscale/resource_dns_configuration_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+			return nil
+		}
+	}
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviderFactories(t),
+		CheckDestroy: checkResourceDestroyed(resourceName, checkProperties(&tailscale.DNSConfiguration{
+			// TODO: we need empty values here because the server currently doesn't `omitempty`.
```
mpminardi: Reminder to adjust / rewrite when we `omitempty` on the server

mpminardi (approved):
LGTM once the server bits are deployed + the omitempty stuff is addressed

tailscale/resource_dns_configuration_test.go
oxtoacart: Done.

docs/resources/dns_configuration.md
```diff
@@ -0,0 +1,103 @@
...
+  The dns_configuration resource allows you to manage the complete DNS configuration for your Tailscale network. See https://tailscale.com/kb/1054/dns for more information.
+---
+
+# tailscale_dns_configuration (Resource)
+
+The dns_configuration resource allows you to manage the complete DNS configuration for your Tailscale network. See https://tailscale.com/kb/1054/dns for more information.
+
+~> **Warning:** This resource is in alpha, proceed with caution!
+
+~> **Note:** The Tailscale Terraform provider has multiple resources for managing DNS configuration. This resource  is meant to manage the entirety of a Tailnet's DNS configuration and conflicts with [tailscale_dns_nameservers](dns_nameservers.md), [tailscale_dns_preferences](dns_preferences.md), [tailscale_dns_search_paths](dns_search_paths.md), and [tailscale_dns_split_nameservers](dns_split_nameservers.md). This resource and previously mentioned resources should not be used simultaneously.
```
mzbenami: looks like extra space between "resource" and "is":
```
This resource  is meant
```

docs/resources/dns_configuration.md
```diff
@@ -0,0 +1,103 @@
...
+### Read-Only
+
+- `id` (String) The ID of this resource.
+
+<a id="nestedblock--nameservers"></a>
+### Nested Schema for `nameservers`
+
+Required:
+
+- `address` (String) The nameserver's IPv4 or IPv6 address
```
mzbenami, line 70: FYI, no change needed: I think technically this can be a DoH URL as well and it will work, somewhat. But I see the UI only allows IP addresses and the existing API only documents IP addresses. So I'll make a ticket to remember to update this language when we update the language everywhere else.

templates/resources/dns_configuration.md.tmpl
```diff
@@ -0,0 +1,26 @@
...
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+---
+
+# {{.Name}} ({{.Type}})
+
+{{ .Description | trimspace }}
+
+~> **Warning:** This resource is in alpha, proceed with caution!
+
+~> **Note:** The Tailscale Terraform provider has multiple resources for managing DNS configuration. This resource  is meant to manage the entirety of a Tailnet's DNS configuration and conflicts with [tailscale_dns_nameservers](dns_nameservers.md), [tailscale_dns_preferences](dns_preferences.md), [tailscale_dns_search_paths](dns_search_paths.md), and [tailscale_dns_split_nameservers](dns_split_nameservers.md). This resource and previously mentioned resources should not be used simultaneously.
```
mzbenami: Same note about `This resource  is meant`

docs/resources/dns_configuration.md
```diff
@@ -0,0 +1,103 @@
...
+
+Optional:
+
+- `use_with_exit_node` (Boolean) This name server will continue to be used when an exit node is selected (requires Tailscale v1.88 or later). Defaults to false.
```
mzbenami: I prefer we be consistent with `nameserver` (one word), here and throughout.

docs/resources/dns_configuration.md
```diff
@@ -0,0 +1,103 @@
...
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Optional
+
+- `magic_dns` (Boolean) Whether or not to enable magic DNS. Defaults to true.
+- `nameservers` (Block List) Set the nameservers used by devices on your network to resolve DNS queries. (see [below for nested schema](#nestedblock--nameservers))
```
mzbenami: I think we need to be clearer about how this field works in conjunction with `override_local_dns`.  These are only really used when that boolean is set to true. Otherwise, they are "fallback resolvers". However, in our KB and UI text, we don't mention "fallback resolvers", so probably shouldn't get into it here either. I see in the `override_local_dns` test you reference the "DNS servers", but I think we should also do the reverse. And to be extra clear we can mention the nameservers in `split_dns` _don't_ depend on `override_local_dns`. So, to sum it up, I would prefer:

- `nameservers` text explicitly mentions `override_local_dns`.
- `override_local_dns` text explicitly mentions `nameservers` (instead of implicit "DNS servers")
- `split_dns` text explicitly mentions that it is unaffected by `override_local_dns`.

tailscale/resource_dns_configuration.go
```diff
@@ -0,0 +1,257 @@
...
+		ReadContext:   resourceDNSConfigurationRead,
+		CreateContext: resourceDNSConfigurationCreate,
+		UpdateContext: resourceDNSConfigurationUpdate,
+		DeleteContext: resourceDNSConfigurationDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: map[string]*schema.Schema{
+			"nameservers": {
+				Description: "Set the nameservers used by devices on your network to resolve DNS queries.",
```
mzbenami: Similar comment about mentioning `override_local_dns`.

docs/resources/dns_configuration.md
mzbenami: There are a few places where we have help text like this to update.

oxtoacart: @mzbenami Can you please suggest the precise text? I was just copying what's in the UI.

mzbenami: Sure, I think UI has some other non-text contextual communication that we can't replicate here (location of toggle in relation to global list, split-dns in a separate section), so we need to be more explicit IMO. How about:

`nameservers`: (Block List) Set the nameservers used by devices on your network to resolve DNS queries. `override_local_dns` must also be true to prefer these nameservers over local DNS configuration. (see [below for nested schema](#nestedblock--nameservers))

`override_local_dns`:  (Boolean) When enabled, use the configured DNS servers in `nameservers` to resolve names outside the tailnet. When disabled, devices will prefer their local DNS configuration. Defaults to false.

`split_dns`: (Block List) Set the nameservers used by devices on your network to resolve DNS queries on specific domains (requires Tailscale v1.8 or later). Configuration does not depend on `override_local_dns`. (see [below for nested schema](#nestedblock--split_dns))

Having read the above, it's understandable to have the reaction of "if a populated `nameservers` only works if we set another thing, then what's the point?" I think that's valid, but it's an artifact of the heavy weighting we gave to matching the UI in designing this and the new API. One could make the argument that it also provides convenience for toggling local/global on and off without having to reconfigure the resolver addresses.

oxtoacart: Thank you!

mzbenami (approved):

docs/resources/dns_configuration.md
```diff
@@ -0,0 +1,103 @@
...
+    magic_dns = true
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Optional
+
+- `magic_dns` (Boolean) Whether or not to enable magic DNS. Defaults to true.
```
mzbenami: Sorry, I already approved, but working on OpenAPI docs now and looking at UI and existing API docs, we always stylize to `MagicDNS`, so can you do s/magic DNS/MagicDNS/g.

oxtoacart:
Final [acceptance test](https://github.com/tailscale/corp/actions/runs/17860866494) passed.

PR #554, opened 2025-10-03
tailscale/resource_acl: relax policy validation during plan steps

### LINKED ISSUE #546: Allow bypassing ACL validation for groups
We create our IdP groups via Terraform. If we reference a group in our ACL that is due to be created in the same PR, the provider will error with "ACL validation failed: group not found".

Either the provider should recognise that the group will be created (we enforce the dependency) or provide the option to disable group validation.

mcoulombe:

Some resources like SCIM groups can be created and added as a reference to the policy file in the same run. The eager validation causes otherwise valid runs to fail at the plan step because the references do not exist yet.

We are relaxing the plan validation logic to avoid such false positives. The full validation results are still available in debug logs in case this is useful. Note that the validation is done in full during an apply before the policy file is changed so this does not permit a tailnet policy to be in an invalid state.

Fixes #546

Example of a validation error for a failed test in the debug logs:
```
2025-10-03T14:07:54.620-0400 [DEBUG] provider.terraform-provider-tailscale: ACL validation unsuccessful due to advisory error: tf_resource_type=tailscale_acl tf_rpc=PlanResourceChange @module=provider tf_provider_addr=provider error="ACL validation failed: test(s) failed; [{insecure@example.com [[acl test error]: address \"1.2.3.4:80\" (protocol \"tcp\"): want: Drop, got: Accept]}]" tf_req_id=4a4b79ad-9fb5-fe55-1ab8-961f4e20b4af @caller=/Users/maxc/Documents/projects/terraform-provider-tailscale/tailscale/resource_acl.go:54 timestamp=2025-10-03T14:07:54.620-0400
```

Resulting apply if the user ignores or does not check the validation logs before an apply:
```
Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

tailscale_acl.as_hujson: Modifying... [id=acl]
╷
│ Error: Failed to set ACL
│ 
│   with tailscale_acl.as_hujson,
│   on main.tf line 15, in resource "tailscale_acl" "as_hujson":
│   15: resource "tailscale_acl" "as_hujson" {
│ 
│ test(s) failed (400)
╵
╷
│ Error: user: insecure@example.com
│ error: [acl test error]: address "1.2.3.4:80" (protocol "tcp"): want: Drop, got: Accept
│ 
│   with tailscale_acl.as_hujson,
│   on main.tf line 15, in resource "tailscale_acl" "as_hujson":
│   15: resource "tailscale_acl" "as_hujson" {
│ 
╵

```


docs/resources/acl.md
```diff
@@ -33,12 +39,12 @@ resource "tailscale_acl" "as_hujson" {
...
-    "acls": [
+    "grants": [
       {
         // Allow all users access to all ports.
-        action = "accept",
-        users  = ["*"],
-        ports  = ["*:*"],
+        "src": ["*"],
+        "dst": ["*"],
+        "ip":  ["*"],
```
mcoulombe: Drive-by, the HuJSON example was not valid. Switching from the ACL to the Grant syntax at the same time.

tailscale/resource_acl.go
```diff
@@ -42,7 +46,17 @@ func resourceACL() *schema.Resource {
...
 			}
-			return client.PolicyFile().Validate(ctx, rd.Get("acl").(string))
+			err := client.PolicyFile().Validate(ctx, rd.Get("acl").(string))
+			if err != nil && syntaxErrorMessage.MatchString(err.Error()) {
+				return err
+			} else if err != nil {
+				// non-syntax errors are logged but do not fail the plan operation
+				tflog.Debug(ctx, "ACL validation unsuccessful due to advisory error", map[string]interface{}{"error": err})
+				return nil
+			}
```
mcoulombe, line 56: From our conversations, warnings might be counterproductive and confuse people, but we should not make the information completely unavailable.

tailscale/resource_acl.go
```diff
@@ -42,7 +46,17 @@ func resourceACL() *schema.Resource {
 			if rd.Get("acl").(string) == "" {
 				return nil
 			}
-			return client.PolicyFile().Validate(ctx, rd.Get("acl").(string))
+			err := client.PolicyFile().Validate(ctx, rd.Get("acl").(string))
+			if err != nil && syntaxErrorMessage.MatchString(err.Error()) {
+				return err
+			} else if err != nil {
+				// non-syntax errors are logged but do not fail the plan operation
+				tflog.Debug(ctx, "ACL validation unsuccessful due to advisory error", map[string]interface{}{"error": err})
```
oxtoacart, line 54:
```suggestion
				tflog.Debug(ctx, "ACL validation unsuccessful due to advisory warning, apply may fail if dependencies are unmet", map[string]interface{}{"error": err})
```

tailscale/resource_acl.go
```diff
@@ -25,6 +27,8 @@ If tests are defined in the ACL (the top-level "tests" section), ACL validation
 // TODO: use an exported variable when https://github.com/hashicorp/terraform-plugin-sdk/issues/803 has been addressed.
 const UnknownVariableValue = "74D93920-ED26-11E3-AC10-0800200C9A66"
 
+var syntaxErrorMessage = regexp.MustCompile(`^ACL validation failed: line \d+, column \d+:`)
```
oxtoacart, line 30: Maybe include a comment about what assumptions this makes about the control server.

tailscale/resource_acl.go
mcoulombe: I can soften the wording and call them warnings, but the validation failures are not necessarily undefined dependencies. The provider should not characterize what caused the error if it just forwards what the BE returned, which can change in the future.

tailscale/resource_acl.go
mcoulombe: What do you mean by assumptions about the control server?

tailscale/resource_acl.go
oxtoacart: > are not necessarily

Hence the use of "may"

tailscale/resource_acl.go
oxtoacart: The assumption that all syntax errors present as an error in this form, and that if this change, the logic breaks.

Issue #514, opened 2025-06-03
FR: Add grants support


ivant:
**Is your feature request related to a problem? Please describe.**
Tailscale just announced [Grants GA](https://tailscale.com/blog/grants-ga).

**Describe the solution you'd like**
A resource for setting them up would be useful.

mpminardi:
Hey folks! Apologies for the delayed response on this. To clear up some potential confusion: the existing [tailscale_acl](https://registry.terraform.io/providers/tailscale/tailscale/latest/docs/resources/acl) controls management of the _entire_[ policy file](https://tailscale.com/kb/1337/policy-syntax) and _not_ just the `acls` section within it.

For example, the minimal `as_hujson` example resource [in the docs ](https://registry.terraform.io/providers/tailscale/tailscale/latest/docs/resources/acl#example-usage)can be re-written as something like the following today using grants syntax:

```
resource "tailscale_acl" "as_hujson" {
  acl = <<EOF
  {
    // Comments in HuJSON policy are preserved when the policy is applied.
    "grants": [
      // Allow all users access to all ports.
      {"src": ["*"], "dst": ["*"], "ip": ["*"]},
    ],
  }
  EOF
}
```

Is something like the above what people were hoping to be able to do with the provider, or am I misunderstanding the ask here?

nint8835:
Not sure how I missed that fact previously, I just tried doing it that way and can confirm it worked for my needs

mezuzza:
This simply didn't work at the time this ticket was open. tailscale-client-go would fail to parse the json and we would get an error. I don't quite remember the error.

From my understanding, [this change](https://github.com/tailscale/terraform-provider-tailscale/pull/547) fixed this issue since it upgraded from `tailscale.com/client/tailscale/v2 v2.0.0-20250616133344-8dcb33eb281b` to `tailscale.com/client/tailscale/v2 v2.0.0-20250917165625-03d1dde3a1fc`.

You'll note that before that change, the code being ran can be [found here](https://github.com/tailscale/tailscale-client-go-v2/blob/8dcb33eb281bfe7254016318fe5954939600e05b/policyfile.go#L58). The code for after the change is [found here](https://github.com/tailscale/tailscale-client-go-v2/blob/03d1dde3a1fcf3a39596dc9982cd1987181eef9c/policyfile.go#L59).

See how Grants and IPSets was added between those two commits.

PR #564, opened 2025-10-14
all: update docs for ACL resource and data source

mpminardi:
Update the documentation for the ACL resource to clarify that it controls the entire policy file for a tailnet and not just the ACLs section of the policy file.

Update the example usage to show usage of grants over the deprecated acl syntax.

Updates https://github.com/tailscale/terraform-provider-tailscale/issues/514


mcoulombe (approved):
Nice. Food for thought, what's our opinion on replacing the `tailscale_acl` resource/source altogether with `tailscale_policy_file`?

Iiuc the same resource implementation can be added multiple times in the provider's ResourceMap, effectively making aliases. So we could mark the entire `tailscale_acl` as deprecated and have new projects use `tailscale_policy_file`.

Issue #485, opened 2025-03-03
Support using OIDC to authenticate provider


tcaddy:
**Is your feature request related to a problem? Please describe.**
I don't like that I have to provide an API key or a set of OAuth credentials in order to auth the terraform provider.   I'd like to be able to auth in a manner that doesn't require using secrets (which could be used elsewhere to gain access to Tailscale's API).

**Describe the solution you'd like**
I would like you to support OIDC authentication for the Tailscale provider.

**Additional context**
The terraform providers for GCP and Azure AD both support auth'ing with OIDC and it's really nice b/c

1. no secrets are involved
2. there are no long lived credentials used
3. there is no need to reset an Oauth client every 30/60/90/etc days
4. it's more secure

casperbiering:
Just found that it's already reported here: https://github.com/tailscale/tailscale/issues/13087

tcaddy:
Oof, sorry to make a dupe issue.   I searched for "OIDC" before posting this.  That dupe issue is from another git repo so I think maybe this one should stay open since the change would happen in this repo, right?

PR #567, opened 2025-10-15
provider: added workload identity federation auth support

mcoulombe:

Allows the Tailscale Terraform provider to dynamically generate an API access token by exchanging an identity token. This way providers do not need to be configured with sensitive information like an authkey or oauth_client_secret.

Configuring a federated identity OAuth client is private at the time of this PR but public access is coming soon™️. 

Updates #485

Depends on this PR on the client. I'll update the dependency with the proper stable release before merging.


tailscale/provider.go
```diff
@@ -42,24 +42,31 @@ func Provider(options ...ProviderOption) *schema.Provider {
...
-				Description: "The OAuth application's secret when using OAuth client credentials. Can be set via the TAILSCALE_OAUTH_CLIENT_SECRET environment variable. Both 'oauth_client_id' and 'oauth_client_secret' must be set. Conflicts with 'api_key'.",
+				Description: "The OAuth application's secret when using OAuth client credentials. Can be set via the TAILSCALE_OAUTH_CLIENT_SECRET environment variable. Conflicts with 'api_key' and 'identity_token'.",
 				Sensitive:   true,
 			},
 			"scopes": {
 				Type:        schema.TypeList,
 				Optional:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
-				Description: "The OAuth 2.0 scopes to request when for the access token generated using the supplied OAuth client credentials. See https://tailscale.com/kb/1215/oauth-clients/#scopes for available scopes. Only valid when both 'oauth_client_id' and 'oauth_client_secret' are set.",
+				Description: "The OAuth 2.0 scopes to request for when generating the access token using the supplied OAuth client credentials. See https://tailscale.com/kb/1215/oauth-clients/#scopes for available scopes. Only valid when both 'oauth_client_id' and 'oauth_client_secret' are set.",
```
oxtoacart:
```suggestion
				Description: "The OAuth 2.0 scopes to request when generating the access token using the supplied OAuth client credentials. See https://tailscale.com/kb/1215/oauth-clients/#scopes for available scopes. Only valid when both 'oauth_client_id' and 'oauth_client_secret' are set.",
```

tailscale/provider.go
```diff
@@ -42,24 +42,31 @@ func Provider(options ...ProviderOption) *schema.Provider {
 				Description: "The API key to use for authenticating requests to the API. Can be set via the TAILSCALE_API_KEY environment variable. Conflicts with 'oauth_client_id' and 'oauth_client_secret'.",
 				Sensitive:   true,
 			},
+			"identity_token": {
```
oxtoacart, line 46: Do you want to name this and the environment variable something that makes it clear it's for workload identity federation?

mcoulombe: I don't have a strong preference, do you mean `wif_identity_token`? @mpminardi any thoughts on the naming?

mcoulombe: Follow-up from a Slack conversation, Sam felt against the wif prefix and we'll support both TAILSCALE_IDENTITY_TOKEN and IDENTITY_TOKEN env vars to be consistent with the OAuth fields.

oxtoacart (approved):
Still LGTM.

PR #609, opened 2026-01-08
examples: fix JSON and HuJSON ACL syntax

alisdair:
The ACL example includes mixed syntax for the JSON case, and invalid syntax for the HuJSON case. This commit fixes both.

### DIFF
--- a/examples/resources/tailscale_acl/resource.tf
+++ b/examples/resources/tailscale_acl/resource.tf
@@ -1,6 +1,6 @@
 resource "tailscale_acl" "as_json" {
   acl = jsonencode({
-    grants : [
+    grants = [
       {
         // Allow all users access to all ports.
         src = ["*"],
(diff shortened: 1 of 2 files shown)


examples/resources/tailscale_acl/resource.tf
```diff
@@ -18,9 +18,9 @@ resource "tailscale_acl" "as_hujson" {
     "grants": [
       {
         // Allow all users access to all ports.
-        "src" = ["*"],
-        "dst" = ["*"],
-        "ip"  = ["*"],
+        "src":["*"],
```
sfllaw: ✍️ Style (non-blocking): Missing space?
```suggestion
        "src": ["*"],
```

alisdair: Good eye, thanks!

PR #617, opened 2026-01-23
tailscale: add gcs log streaming

danni-popova:
Add configuration options for GCS log streaming.

Updates https://github.com/tailscale/corp/issues/33335


tailscale/resource_logstream_configuration.go
```diff
@@ -159,6 +182,10 @@ func resourceLogstreamConfigurationCreate(ctx context.Context, d *schema.Resourc
 	s3SecretAccessKey := d.Get("s3_secret_access_key").(string)
 	s3RoleARN := d.Get("s3_role_arn").(string)
 	s3ExternalID := d.Get("s3_external_id").(string)
+	gcsCredentials := d.Get("gcs_credentials").(string)
+	gcsScopes := d.Get("gcs_scopes").([]string)
```
mpminardi: We'll need to do some extra work here to convert between the Terraform representation -> the API representation here. Something like:
```suggestion
	var gcsScopes []string
	for _, scope := range d.Get("gcs_scopes").(*schema.Set).List() {
		gcsScopes = append(gcsScopes, scope.(string))
	}
```

tailscale/resource_logstream_configuration_test.go
```diff
@@ -67,6 +67,18 @@ const testLogstreamConfigurationUpdateS3AccessKey = `
 		compression_format     = "zstd"
 	}`
 
+const testLogstreamConfigurationGCS = `
+	resource "tailscale_logstream_configuration" "test_logstream_configuration" {
+		log_type               = "network"
+		destination_type       = "gcs"
```
mpminardi, line 74: Looks like this is missing in the `ValidateFunc` for `destination_type`

tailscale/resource_logstream_configuration_test.go
```diff
@@ -67,6 +67,18 @@ const testLogstreamConfigurationUpdateS3AccessKey = `
...
+		gcs_bucket             = "example-bucket"
+		gcs_scopes             = ["scope1", "scope2"]
+		gcs_credentals         = "{"some":"credentials"}"
```
mpminardi:
```suggestion
		gcs_credentials         = "{"some":"credentials"}"
```

Does this also need to be in the shape of the credentials file to have the backend be happy with it?

tailscale/resource_logstream_configuration_test.go
```diff
@@ -67,6 +67,18 @@ const testLogstreamConfigurationUpdateS3AccessKey = `
...
+		gcs_credentals         = "some-prefix"
```
mpminardi: Is this meant to be:

```suggestion
		gcs_key_prefix         = "some-prefix"
```

potentially?

danni-popova: yes and I misspelled credentials ...

Issue #626, opened 2026-02-06
Add support for filter functionality to devices data source


zbuchheit:
**Is your feature request related to a problem? Please describe.**
Recently filter functionality has been added for devices to the go provider, it would be nice to expose this functionality in this provider.

See: https://github.com/tailscale/tailscale-client-go-v2/pull/82

**Describe the solution you'd like**
I would like to be able to filter results when using the devices data source.

**Additional context**

PR #631, opened 2026-02-10
data_source_devices: add filters

neinkeinkaffee:
This adds support for device list filtering. The filter interface takes inspiration from e.g. the `aws_ami` data source in the AWS Terraform provider.

Usage:
```
data "tailscale_devices" "all_devices" { 
	filter { 
		name = "isEphemeral"
		values = ["true"]
	}
	filter { 
		name = "tags"
		values = ["tag:server", "tag:test"]
	}
}
```

Updates #626

### DIFF
--- a/docs/data-sources/devices.md
+++ b/docs/data-sources/devices.md
@@ -15,6 +15,15 @@ The devices data source describes a list of devices in a tailnet
 ```terraform
 data "tailscale_devices" "sample_devices" {
   name_prefix = "example-"
+
+  filter {
+    name = "isEphemeral"
+    values = ["true"]
+  }
+  filter {
+    name = "tags"
+    values = ["tag:server", "tag:test"]
+  }
 }
 ```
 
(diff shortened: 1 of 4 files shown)


tailscale/data_source_devices_test.go
```diff
@@ -134,4 +117,85 @@ func TestAccTailscaleDevices(t *testing.T) {
 			},
 		},
 	})
+
+	// Test tailscale_devices with filter applied.
+	resource.Test(t, resource.TestCase{
```
neinkeinkaffee, line 123: This copies the existing test for the data source and applies filters (except that the existing test also creates and validates data sources for individual data sources, which we're not repeating here). I've extracted the logic that asserts that the devices returned by the data source and the client have equal attributes.

tailscale/data_source_devices_test.go
```diff
@@ -134,4 +117,85 @@ func TestAccTailscaleDevices(t *testing.T) {
...
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: `data "tailscale_devices" "all_devices" {
```
mpminardi: Very minor nit:

```suggestion
				Config: `data "tailscale_devices" "filtered_devices" { 
```

docs/data-sources/devices.md
```diff
@@ -23,13 +23,23 @@ data "tailscale_devices" "sample_devices" {
 
 ### Optional
 
+- `filter` (Block Set) Filters the device list to elements devices whose fields match the provided values. (see [below for nested schema](#nestedblock--filter))
```
mpminardi, line 35: It might also be good to show an example of using filters in the "Example usage" above! That would involve modifying https://github.com/tailscale/terraform-provider-tailscale/blob/main/examples/data-sources/tailscale_devices/data-source.tf and then doing a generate again

neinkeinkaffee: Nice, the existence of those had escaped me!

PR #645, opened 2026-03-06
all: create a Mux terraform for Terraform SDKv2 + plugin framework

alexwlchan:
Start migrating to the Terraform plugin framework by [serving the provider][1] using both the existing SDKv2 and the framework served via [a mux][2]. This patch contains the basic mux server and enough changes to get the acceptance tests passing, but doesn't actually implement any resources or data sources.

[1]: https://developer.hashicorp.com/terraform/plugin/framework/migrating/providers#serving-the-provider
[2]: https://developer.hashicorp.com/terraform/plugin/framework/migrating/mux

Updates tailscale/corp#37220

### DIFF
--- a/tailscale/provider_test.go
+++ b/tailscale/provider_test.go
@@ -12,6 +12,10 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/hashicorp/terraform-plugin-framework/providerserver"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
+	"github.com/hashicorp/terraform-plugin-mux/tf5muxserver"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -51,12 +55,29 @@ func testAccPreCheck(t *testing.T) {
 	}
 }
 
-func testAccProviderFactories(t *testing.T) map[string]func() (*schema.Provider, error) {
+// testAccProviderFactories creates a mux server that serves both the SDKv2 and
+// the plugin framework providers.
+//
+// See https://developer.hashicorp.com/terraform/plugin/framework/migrating/mux#protocol-version-5
+func testAccProviderFactories(t *testing.T) map[string]func() (tfprotov5.ProviderServer, error) {
 	t.Helper()
 
-	return map[string]func() (*schema.Provider, error){
-		"tailscale": func() (*schema.Provider, error) {
-			return Provider(), nil
+	return map[string]func() (tfprotov5.ProviderServer, error){
+		"tailscale": func() (tfprotov5.ProviderServer, error) {
+			ctx := context.Background()
+
+			providers := []func() tfprotov5.ProviderServer{
+				providerserver.NewProtocol5(NewFrameworkProvider()),
+				Provider().GRPCProvider,
+			}
+
+			muxServer, err := tf5muxserver.NewMuxServer(ctx, providers...)
+
+			if err != nil {
+				return nil, err
+			}
+
+			return muxServer.ProviderServer(), nil
 		},
 	}
 }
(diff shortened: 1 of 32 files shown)


tailscale/provider_test.go
```diff
@@ -69,23 +90,40 @@ func TestProvider(t *testing.T) {
...
+			}
+
+			muxServer, err := tf5muxserver.NewMuxServer(ctx, providers...)
+
+			if err != nil {
+				return nil, err
+			}
 
-				// Don't require any of the global configuration
-				p.Schema = nil
```
alexwlchan, line 87: Because Terraform checks that the schema is consistent across the muxed providers, our old strategy of deleting the schema for the tests doesn't
work; instead, we set an API key via environment variable. The `t.Setenv` helper can't be used in parallel, so we have to switch parallel tests to
serial.

tailscale/provider.go
```diff
@@ -27,7 +32,202 @@ var providerVersion = "dev"
 
 type ProviderOption func(p *schema.Provider)
 
-// Provider returns the *schema.Provider instance that implements the terraform provider.
+type tailscaleProvider struct {
```
mpminardi: Nit: might be better to have the framework specific logic live in a `provider_framework.go` or similar to more easily tell between the different providers at a glance.

tailscale/provider.go
```diff
@@ -27,7 +32,202 @@ var providerVersion = "dev"
...
+	Client tailscale.Client
+}
+
+func New() provider.Provider {
+	return &tailscaleProvider{}
+}
+
+// Metadata defines information about the provider itself.
+func (p *tailscaleProvider) Metadata(ctx context.Context, req provider.MetadataRequest, resp *provider.MetadataResponse) {
```
mpminardi: Nit:

```suggestion
func (p *tailscaleProvider) Metadata(_ context.Context, _ provider.MetadataRequest, resp *provider.MetadataResponse) {
```

tailscale/provider.go
```diff
@@ -27,7 +32,202 @@ var providerVersion = "dev"
...
+	resp.TypeName = "tailscale"
+	resp.Version = providerVersion
+}
+
+// Schema defines a [schemav2.Schema] describing what data is available in the provider's
+// configuration.
+func (p *tailscaleProvider) Schema(ctx context.Context, req provider.SchemaRequest, resp *provider.SchemaResponse) {
```
mpminardi: Nit:

```suggestion
func (p *tailscaleProvider) Schema(_ context.Context, _ provider.SchemaRequest, resp *provider.SchemaResponse) {
```

tailscale/provider.go
```diff
@@ -27,7 +32,202 @@ var providerVersion = "dev"
...
+		resp.Diagnostics.AddError(
+			"Could not parse base URL",
+			fmt.Sprintf("While configuring the provider, "+
+				"the base URL %q could not be parsed: %v", baseURL, err),
+		)
+	}
+
+	// TODO(alexc): I copied this check from the old provider, but if I've read the code
+	// correctly it will never be triggered: the default value for tailnet is a hyphen,
+	// not an empty string.
```
mpminardi: I believe this would trigger specifically when the `tailnet` argument is provided as an empty string. E.g., if you have a configuration like:

```hcl
provider "tailscale" {
  api_key = "tskey-api-..."
  tailnet = ""
}
```

vs. the default of `-` applying if `tailnet` is omit entirely or unset as an environment variable.

tailscale/provider.go
```diff
@@ -27,7 +32,202 @@ var providerVersion = "dev"
...
+			},
+		}
+	} else {
+		p.Client = tailscale.Client{
+			BaseURL:   parsedBaseURL,
+			UserAgent: userAgent,
+			APIKey:    apiKey,
+			Tailnet:   tailnet,
+		}
+	}
```
mpminardi: Nit: this is a problem in the original provider, but potentially we can setup the common bits outside of the branching logic to reduce duplicated logic.

E.g., something like:

```go
	p.Client = tailscale.Client{
		BaseURL:   parsedBaseURL,
		UserAgent: userAgent,
		Tailnet:   tailnet,
	}

	if oauthClientID != "" && oauthClientSecret != "" {
		p.Client.Auth = &tailscale.OAuth{
			ClientID:     oauthClientID,
			ClientSecret: oauthClientSecret,
			Scopes:       scopes,
		}
	} else if oauthClientID != "" && identityToken != "" {
		p.Client.Auth = &tailscale.IdentityFederation{
			ClientID: oauthClientID,
			IDTokenFunc: func() (string, error) {
				return identityToken, nil
			},
		}
	} else {
		p.Client.APIKey = apiKey
	}
```

tailscale/provider.go
```diff
@@ -27,7 +32,202 @@ var providerVersion = "dev"
 
 type ProviderOption func(p *schema.Provider)
 
-// Provider returns the *schema.Provider instance that implements the terraform provider.
+type tailscaleProvider struct {
+	Client tailscale.Client
+}
+
+func New() provider.Provider {
```
mpminardi: Nit: maybe `NewFrameworkProvider`?

tailscale/provider_test.go
mpminardi: I think this was originally done for our "unit tests" (e.g., those marked with `IsUnitTest: true` in `resource.TestCase`) as they don't run against a real server but do use the Terraform acceptance testing logic.

I think we can likely remove most (if not all) tests of this form as a followup / cleanup action since they don't provide much actual value vs. acceptance tests. E.g., they are a very bare minimum check that a plan / apply can succeed against a completely mocked out server that does not actually return data in the correct shape.

alexwlchan: I'm going to leave that for a separate PR, rather than stack new changes into this patch.

One reason we might want to keep them: AIUI we don't currently run acceptance tests against PRs? So those minimal tests give us some safety net that something isn't completely broken when we do small patches (e.g. adding Fleet and Huntress last week).

mpminardi: That sounds good to me! Yeah we don't currently have an automated way of running acceptance tests against PRs aside from the manual workflow trigger in corp. 

Something I've wanted for a bit but not had the time to explore further is adding a GitHub app / bot that listens for PRs here, downloads the repo at the branch, and runs the acceptance test flows (assuming that the author is a contributor or that we have explicitly approved running against a branch from a fork). This is an even further in the future thing, but would be a lot nicer UX for PRs here if / when we have it 🙂

alexwlchan: Created as a separate ticket: https://github.com/tailscale/terraform-provider-tailscale/issues/649

PR #646, opened 2026-03-06
data_source_acl: migrate to plugin framework

alexwlchan:
This passes all the acceptance tests with `make testacc_local`.

Fixes https://github.com/tailscale/corp/issues/37223


tailscale/data_source_acl.go
```diff
@@ -6,53 +6,77 @@ package tailscale
...
+	data := aclDataSourceModel{
+		ID:     types.StringValue(createUUID()),
+		HuJSON: types.StringValue(huj.String()),
+
+		// TODO(alexc): I copied this from the SDKv2 code when upgrading to the
+		// plugin framework, but I'm not sure it's correct.
+		//
+		// Reading the description of this field, I expect it to return a valid
+		// JSON string, but this could be HuJSON which is invalid JSON.
+		JSON: types.StringValue(huj.String()),
```
mpminardi: We'll want something like:

```go
hujsonString := huj.String()
    
// Minimize transforms the underlying HuJSON representation into valid JSON.
// This is an in-place change of the [hujson.Value] and must be done after we have
// stored the value of the HuJSON representation.
huj.Minimize()
jsonString := huj.String()

data := aclDataSourceModel{
    ID:     types.StringValue(createUUID()),
	HuJSON: types.StringValue(hujsonString),
	JSON: types.StringValue(jsonString),
}
```

here as your comment is indeed correct that the value could otherwise be HuJSON. It looks like the original `huj.Minimize()` got dropped in the shuffle.

tailscale/data_source_acl.go
```diff
@@ -6,53 +6,77 @@ package tailscale
...
 	"github.com/tailscale/hujson"
 )
 
-func dataSourceACL() *schema.Resource {
-	return &schema.Resource{
-		Description: "The acl data source gets the Tailscale policy file for a tailnet",
-		ReadContext: dataSourceACLRead,
-		Schema: map[string]*schema.Schema{
-			"json": {
+// NewACLDataResource returns a new ACL data source.
```
mpminardi: Nit:

```suggestion
// NewACLDataSource returns a new ACL data source.
```

tailscale/data_source_acl.go
alexwlchan: Okay, I'll fix it and add a test.

docs/data-sources/acl.md
```diff
@@ -2,20 +2,26 @@
...
+Returns the Tailscale policy file for a tailnet.
 
 ~> **Note:** The naming of this data source predates Tailscale's usage of the term "policy file" to refer to the centralized configuration file for a tailnet. This data source fetches a tailnet's entire policy file and not just the ACLs section within it.
 
+## Example Usage
+
+```terraform
+data "tailscale_acl" "example" {}
+```
+
```
mpminardi, line 19: Ah good catch on this!

mpminardi (approved):

Issue #644, opened 2026-03-05
Acceptance tests are flaky when comparing timestamps


alexwlchan:
In particular there are tests that rely on second-level precision which sometimes fail:

> data_source_devices_test.go:87: Step 1/1 error: Check failed: wrong value for property last_seen of user data.tailscale_device.flamingo_ostrich_burgandy_sparrow_gold_mauve_donkey, want 2026-03-05T15:31:05Z, got 2026-03-05T15:31:04Z

PR #663, opened 2026-03-26
data_source_device: migrate to plugin framework

neinkeinkaffee:
This lifts and shifts the data_source_device from SDKv2 to the Plugin Framework.

Terraform acceptance test run: https://github.com/tailscale/corp/actions/runs/23590973488/job/68695835633

Fixes tailscale/corp#37224

### DIFF
--- a/tailscale/data_source_device.go
+++ b/tailscale/data_source_device.go
@@ -16,6 +16,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"tailscale.com/client/tailscale/v2"
+	"tailscale.com/types/bools"
 )
 
 type deviceDataSourceModel struct {
@@ -287,7 +288,16 @@ func toDeviceDataSourceModel(ctx context.Context, device *tailscale.Device) (dev
 	}
 	data.Addresses = addresses
 
-	tags, diagnostics := types.SetValueFrom(ctx, types.StringType, device.Tags)
+	// Normalize nil to empty slice before converting it to plugin framework
+	// types.
+	// This is to avoid drift when upgrading from a provider version that
+	// is still backed by SDKv2.
+	deviceTags := bools.IfElse(device.Tags != nil, device.Tags, []string{})
+
+	tags, diagnostics := types.SetValueFrom(ctx, types.StringType, deviceTags)
+	if diagnostics.HasError() {
+		return deviceDataSourceModel{}, diagnostics
+	}
 	data.Tags = tags
 
 	return data, diag.Diagnostics{}


tailscale/data_source_device.go
```diff
@@ -8,185 +8,298 @@ import (
...
+		return deviceDataSourceModel{}, diagnostics
+	}
+	data.Addresses = addresses
+
+	tags, diagnostics := types.SetValueFrom(ctx, types.StringType, device.Tags)
+	if diagnostics.HasError() {
+		return deviceDataSourceModel{}, diagnostics
+	}
+	// Replace empty tags by explicit empty list to avoid phantom drift
+	// in the course of migrating to the Terraform Plugin Framework.
```
neinkeinkaffee: There might be a better way? This is to prevent drift along the lines of ` [] -> null`.

In context:
```
~ resource "terraform_data" "the_external_machine" {
                id     = "16f6cd6f-1e88-69cf-bab7-c10e236ed1c0"
              ~ input  = {
                    id                          = "4675775454370919"
                    name                        = "the-external-machine.tailb1ca.ts.net"
                  ~ tags                        = [] -> null
                    # (19 unchanged attributes hidden)
                }
              ~ output = {
                  - addresses                   = [
                      - "100.119.92.40",
                      - "fd7a:115c:a1e0::8501:5c28",
                    ]
                  - authorized                  = true
                  - blocks_incoming_connections = false
                  - client_version              = ""
                  - created                     = "0001-01-01T00:00:00Z"
                  - expires                     = "2026-09-21T18:04:03Z"
                  - hostname                    = "the external machine"
                  - id                          = "4675775454370919"
                  - is_external                 = true
                  - key_expiry_disabled         = false
                  - last_seen                   = "2026-03-26T00:36:24Z"
                  - machine_key                 = ""
                  - name                        = "the-external-machine.tailb1ca.ts.net"
                  - node_id                     = "nxwnTAhfWd11DEVEL"
                  - node_key                    = "nodekey:2bb4583fdc788272b56b6f683ff5220abf814d104666a453a01bb29985375c27"
                  - os                          = "iOS"
                  - tags                        = []
                  - tailnet_lock_error          = ""
                  - tailnet_lock_key            = ""
                  - update_available            = false
                  - user                        = "test@external.com"
                  - wait_for                    = null
                } -> (known after apply)
            }

[...]
```

tailscale/data_source_devices_test.go
```diff
@@ -66,13 +70,15 @@ func TestAccTailscaleDevices(t *testing.T) {
...
 					assertDeviceAttributesEqual(t, rs, deviceIndexes, devicesByID)
 
-					// Now set up device datasources for each device. This is used in the following test
-					// of the tailscale_device datasource.
+					// Now set up a config with a datasource and terraform_data reference for each device.
+					// This is used in the following tests of the tailscale_device datasource.
 					for _, device := range devices {
 						if device.Hostname != "" {
 							devicesDataSources.WriteString(fmt.Sprintf("\ndata \"tailscale_device\" \"%s\" {\n  hostname = \"%s\"\n}\n", toResourceComponent(device.Hostname), device.Hostname))
+							devicesTerraformData.WriteString(fmt.Sprintf("\nresource \"terraform_data\" \"%s\" {\n  input = data.tailscale_device.%s\n  lifecycle {ignore_changes = [ input.last_seen ]}\n}\n", toResourceComponent(device.Hostname), toResourceComponent(device.Hostname)))
```
neinkeinkaffee, line 78: Ignoring timestamp changes until we've tackled #644.

tailscale/data_source_device.go
mpminardi: Did you see this when running the acceptance tests? Running this locally with this commented out seems to pass for me at the least.

neinkeinkaffee: I removed it and run a few times - I run into this about 80%-90% of the time locally. I'm happy to remove the workaround for now and dig a bit deeper what's causing it.

neinkeinkaffee: Failing (non-local) test run for reference: https://github.com/tailscale/corp/actions/runs/23636068897/job/68845495253

neinkeinkaffee: What makes me scratch my head is why this is non-deterministic.

neinkeinkaffee: I'll think a bit more about this but I think the best solution I can think of so far might be to accept that this drift is (sometimes) occurring for SDKv2 -> Plugin Framework and just ignore the tags field in the compatibility test...

neinkeinkaffee: Having thought a bit more about it, I've realised keeping the drift potential on migration in there isn't great because the tiny `[] -> null` in the datasource could snowball into a bigger change and while maybe not super disruptive it's just unnecessary and confusing. I've now settled on "normalising" the tags in the client response to an empty slice before converting to the framework set type: https://github.com/tailscale/terraform-provider-tailscale/pull/663/changes/904ff141e38b994e28f38569d0f0b4715ee877d4

What do you think?

mpminardi (approved):

PR #667, opened 2026-03-27
resource_device_key: migrate to plugin framework

neinkeinkaffee:
This migrates the device key resource to the plugin framework.

This also introduces a RequiresReplace plan modifier to the device ID attribute. The device key's resource ID currently tracks the device ID. 

Previously it was possible for to change the device ID in a config to point to a different device, but the underlying resource ID would keep pointing to the original device, which can quickly result in situations with inconsistent state. 

For example, I could apply a device key resource, then update the device_id. Whenever I now re-apply, the plan will show the device_id changing to the new device ID but the device that gets changed is the device with the original device ID. 

Introducing RequiresReplace for the device ID was the most economic way to resolve this.

Terraform acceptance test run: https://github.com/tailscale/corp/actions/runs/23747478031/job/69179916438

Fixes tailscale/corp#37232

### DIFF
--- a/tailscale/resource_device_key.go
+++ b/tailscale/resource_device_key.go
@@ -6,30 +6,49 @@ package tailscale
 import (
 	"context"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 	"tailscale.com/client/tailscale/v2"
 )
 
-func resourceDeviceKey() *schema.Resource {
-	return &schema.Resource{
-		Description:   "The device_key resource allows you to update the properties of a device's key",
-		ReadContext:   resourceDeviceKeyRead,
-		CreateContext: resourceDeviceKeyCreate,
-		DeleteContext: resourceDeviceKeyDelete,
-		UpdateContext: resourceDeviceKeyUpdate,
-		Importer: &schema.ResourceImporter{
-			StateContext: schema.ImportStatePassthroughContext,
-		},
-		Schema: map[string]*schema.Schema{
-			"device_id": {
-				Type:        schema.TypeString,
+type deviceKeyResourceModel struct {
+	ID                types.String `tfsdk:"id"`
+	DeviceID          types.String `tfsdk:"device_id"`
+	KeyExpiryDisabled types.Bool   `tfsdk:"key_expiry_disabled"`
+}
+
+// NewDeviceKeyResource returns a new device key resource.
+func NewDeviceKeyResource() resource.Resource {
+	return &deviceKeyResource{}
+}
+
+type deviceKeyResource struct {
+	ResourceBase
+}
+
+func (d deviceKeyResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_device_key"
+}
+
+func (d deviceKeyResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "The device_key resource allows you to update the properties of a device's key",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed: true,
+			},
+			"device_id": schema.StringAttribute{
 				Required:    true,
 				Description: "The device to update the key properties of",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
 			},
-			"key_expiry_disabled": {
-				Type:        schema.TypeBool,
+			"key_expiry_disabled": schema.BoolAttribute{
 				Optional:    true,
 				Description: "Determines whether or not the device's key will expire. Defaults to `false`.",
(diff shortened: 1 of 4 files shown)


tailscale/resource_device_key.go
```diff
@@ -83,29 +127,47 @@ func resourceDeviceKeyRead(ctx context.Context, d *schema.ResourceData, m interf
...
+	if err := d.Client.Devices().SetKey(ctx, deviceID, key); err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to update device key",
+			"Could not update device key: "+err.Error(),
+		)
+		return
 	}
 
-	return resourceDeviceKeyRead(ctx, d, m)
+	plan.ID = types.StringValue(deviceID)
```
neinkeinkaffee, line 162: The ID we get from resource.UpdateRequest.Plan is empty, so we need to set it explicitly to the deviceID whose key this resource manages before saving it in state. 

Theoretically we could copy the ID from resource.UpdateRequest.State, but then we would run into issues if someone changed the `device_id` of an existing device key resource - unless we combine it with "requires replace" semantics, but that's outside the scope of this migration.

tailscale/resource_device_key_test.go
```diff
@@ -101,6 +101,25 @@ func TestAccTailscaleDeviceKey(t *testing.T) {
...
 	})
+
+	// Migration test to ensure the resource is unchanged when migrating
+	// from the plugin SDK to the plugin framework.
+	//
+	// See https://developer.hashicorp.com/terraform/plugin/framework/migrating/testing#terraform-data-resource-example
+	checkResourceIsUnchangedInPluginFramework(t,
+		fmt.Sprintf(testDeviceKeyCreate, os.Getenv("TAILSCALE_TEST_DEVICE_NAME")),
+		resource.ComposeTestCheckFunc(
+			checkResourceRemoteProperties(resourceName, checkProperties(false)),
```
neinkeinkaffee, line 112: I'm of two minds whether the checks to run in the context of the resource is unchanged tests should be all tests we've got or just a subset. @alexwlchan what was your interpretation of what the job of the check we pass here is? Is it to sanity check that the old provider version still works because otherwise we'd not be interested in proceeding to drift detection?

tailscale/resource_device_key.go
```diff
@@ -6,75 +6,119 @@ package tailscale
...
+
+	deviceID := state.ID.ValueString()
 
-	device, err := client.Devices().Get(ctx, deviceID)
+	device, err := d.Client.Devices().Get(ctx, deviceID)
 	if err != nil {
-		return diagnosticsError(err, "Failed to fetch devices")
+		resp.Diagnostics.AddError(
+			"Failed to fetch devices",
+			"Could not read device key with device ID "+deviceID+": "+err.Error(),
```
alexwlchan: nit:

```suggestion
			"Could not read device key for device with ID "+deviceID+": "+err.Error(),
```

tailscale/resource_device_key.go
alexwlchan: This sounds fine; perhaps this comment should be in the source code as a TODO?

tailscale/resource_device_key_test.go
alexwlchan: > Is it to sanity check that the old provider version still works because otherwise we'd not be interested in proceeding to drift detection?

I'm not sure I understand this sentence?

I would run all the tests if we can; it's minimal cost and will help us catch unexpected drift/migration issues.

neinkeinkaffee: Ah, with "check" I was referring to what we pass into `checkResourceIsUnchangedInPluginFramework` via the `check resource.TestCheckFunc` parameter. 

I think they are meant to ensure whether the old provider version has created the resource the way we expect. And if it hasn't, we don't even proceed to the next step which checks drift between the old and new provider. Is that correct?

tailscale/resource_device_key.go
neinkeinkaffee: That could be good way to track it! It will be good to look at all device attribute resource together to make that call. The device_authorization and device_tags also use the deviceID as ID, whereas device_subnet_routes store a UUID instead.

tailscale/resource_device_key_test.go
alexwlchan: Yes, you've understood correctly.

tailscale/resource_device_key.go
neinkeinkaffee: I'm starting to change my mind on this one. I'll compare the ergonomics manually first, but I think it should move to RequiresReplace as part of the migration.

tailscale/resource_device_key.go
```diff
@@ -6,75 +6,124 @@ package tailscale
...
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed: true,
+			},
+			"device_id": schema.StringAttribute{
 				Required:    true,
 				Description: "The device to update the key properties of",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
```
neinkeinkaffee, line 49: When the `device_id` changes in the config, the resource will get recreated with a resource ID that matches the `device_id`. This is the simplest way to avoid inconsistencies resulting from the prior behaviour where the resource ID would remain pointing to the old `device_id`.

tailscale/resource_device_key.go
neinkeinkaffee: Done. I've introduce a RequireReplace plan modifier for the deviceID.

tailscale/resource_device_key.go
```diff
@@ -6,75 +6,124 @@ package tailscale
...
 	}
 
-	d.SetId(deviceID)
-	return resourceDeviceKeyRead(ctx, d, m)
+	plan.ID = types.StringValue(deviceID)
+	diags = resp.State.Set(ctx, plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
neinkeinkaffee: All these 
```
if resp.Diagnostics.HasError() {
	return
}
```
before returning are pointless - I'll make sure to remove them everywhere before merging!

mpminardi (approved):

PR #672, opened 2026-03-31
resource_device_subnet_routes: migrate to plugin framework

neinkeinkaffee:
Updates tailscale/corp#37233

Terraform acceptance test: https://github.com/tailscale/corp/actions/runs/23804397668/job/69373568027

### DIFF
--- a/tailscale/resource_device_subnet_routes.go
+++ b/tailscale/resource_device_subnet_routes.go
@@ -6,9 +6,12 @@ package tailscale
 import (
 	"context"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 	"tailscale.com/client/tailscale/v2"
 )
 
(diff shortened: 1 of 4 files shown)


tailscale/resource_device_subnet_routes.go
```diff
@@ -19,6 +25,196 @@ Routes must be both advertised and enabled for a device to act as a subnet route
...
+	if err := d.Client.Devices().SetSubnetRoutes(ctx, deviceID, []string{}); err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to delete device subnet routes",
+			"Could not delete subnet routes: "+err.Error(),
+		)
+		return
+	}
+}
+
 func resourceDeviceSubnetRoutes() *schema.Resource {
```
alexwlchan: Can we get rid of this function now?

neinkeinkaffee: We can!

tailscale/resource_device_subnet_routes.go
```diff
@@ -19,110 +22,192 @@ Routes must be both advertised and enabled for a device to act as a subnet route
...
+		return
 	}
 
-	return nil
+	deviceID := state.DeviceID.ValueString()
+
+	if err := d.Client.Devices().SetSubnetRoutes(ctx, deviceID, []string{}); err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to delete device subnet routes",
+			"Could not delete subnet routes: "+err.Error(),
```
alexwlchan:
```suggestion
			"Could not delete subnet routes for device with ID "+deviceID+": "+err.Error(),
```

tailscale/resource_device_subnet_routes.go
```diff
@@ -19,110 +22,192 @@ Routes must be both advertised and enabled for a device to act as a subnet route
...
+		resp.Diagnostics.AddError(
+			"Failed to set device subnet routes",
+			"Could not set subnet routes for device with ID "+deviceID+": "+err.Error(),
+		)
+		return
 	}
 
-	d.SetId(createUUID())
-	return resourceDeviceSubnetRoutesRead(ctx, d, m)
+	plan.ID = types.StringValue(createUUID())
```
neinkeinkaffee, line 147: This additional line is the only difference between Create and Update. If we switch this resource to using the deviceID as suggested, it becomes straight-forward to remove that duplication.

alexwlchan:
Ah, I missed that!

tailscale/resource_device_subnet_routes.go
```diff
@@ -19,110 +22,178 @@ Routes must be both advertised and enabled for a device to act as a subnet route
...
+}
+
+func (d deviceSubnetRoutesResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: resourceDeviceSubnetRoutesDescription,
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseNonNullStateForUnknown(),
```
mpminardi: Nit: I think we should be safe to have this as `UseStateForUnknown` instead of `UseNonNullStateForUnknown` since this isn't an optional field.

tailscale/resource_device_subnet_routes.go
mpminardi: Yeah this will be a good one to cleanup when we go to `1.x` of the provider!

tailscale/resource_device_subnet_routes.go
neinkeinkaffee: Indeed, I had originally added it as it seemed like the safer bet but meant to double check whether `UseNonNullStateForUnknown` is neccessary here. Looking at the doc comment it seems to primarily intended for use in nested child attributes...

PR #673, opened 2026-04-01
resource_device_tags: migrate to plugin framework

neinkeinkaffee:
Fixes tailscale/corp#37234

Terraform acceptance test: https://github.com/tailscale/corp/actions/runs/23909288214/job/69726659346

### DIFF
--- a/tailscale/resource_device_tags.go
+++ b/tailscale/resource_device_tags.go
@@ -6,58 +6,82 @@ package tailscale
 import (
 	"context"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-
-	"tailscale.com/client/tailscale/v2"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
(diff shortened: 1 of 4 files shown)


tailscale/resource_device_tags.go
```diff
@@ -6,58 +6,82 @@ package tailscale
...
-	var deleteContext = resourceDeviceTagsDelete
-	if isAcceptanceTesting() {
-		// Tags cannot be removed without reauthorizing the device as a user.
-		// We have no way of doing this during testing.
-		// Because of https://github.com/hashicorp/terraform-plugin-sdk/issues/609,
-		// we also have no way of telling the Terraform acceptance test to not test
-		// resource deletion.
-		// So, as a workaround, we don't actually delete during acceptance tests.
-		deleteContext = schema.NoopContext
-	}
```
neinkeinkaffee, line 25: I've moved this check into the Delete method itself as I wasn't sure whether there is a corresponding NoopContext in the plugin framework. There probably is something like that somewhere, but I don't see a massive downside in just returning early inside the function.

neinkeinkaffee: (With "this check" I'm referring to the check above this comment whether we're in an acceptance test and therefore don't want to delete the resource.)

tailscale/resource_device_tags.go
```diff
@@ -66,40 +90,101 @@ func resourceDeviceTagsRead(ctx context.Context, d *schema.ResourceData, m inter
...
+		resp.Diagnostics.AddError(
+			"Failed to set device tags",
+			"Could not set tags for device with ID "+deviceID+": "+err.Error(),
+		)
+		return
 	}
 
-	d.SetId(deviceID)
-	return nil
+	plan.ID = types.StringValue(deviceID)
```
neinkeinkaffee, line 158: Note: Instead of setting the ID explicitly before saving in state here (and in the other resources that we've already migrated), we could use the UseStateForUnknown plan modifier on the ID field. I think it's worth discussing and deciding on one consistent pattern for this.

PR #685, opened 2026-04-10
resource_dns_split_nameservers: migrate to plugin framework

alexwlchan:
Fixes tailscale/corp#37239

### DIFF
--- a/tailscale/resource_dns_split_nameservers.go
+++ b/tailscale/resource_dns_split_nameservers.go
@@ -6,97 +6,165 @@ package tailscale
 import (
 	"context"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 
 	"tailscale.com/client/tailscale/v2"
 )
 
-func resourceDNSSplitNameservers() *schema.Resource {
-	return &schema.Resource{
-		Description:   "The dns_split_nameservers resource allows you to configure split DNS nameservers for your Tailscale network. See https://tailscale.com/kb/1054/dns for more information.",
-		ReadContext:   resourceSplitDNSNameserversRead,
-		CreateContext: resourceSplitDNSNameserversCreateOrUpdate,
-		UpdateContext: resourceSplitDNSNameserversCreateOrUpdate,
-		DeleteContext: resourceSplitDNSNameserversDelete,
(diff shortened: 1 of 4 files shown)


tailscale/resource_dns_split_nameservers.go
```diff
@@ -6,97 +6,165 @@ package tailscale
 import (
 	"context"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	rschema "github.com/hashicorp/terraform-plugin-framework/resource/schema"
```
neinkeinkaffee: Is this still needed?

tailscale/resource_dns_split_nameservers.go
```diff
@@ -6,97 +6,165 @@ package tailscale
...
-		return diagnosticsError(err, "Failed to set dns split nameservers")
+	updateReq := tailscale.SplitDNSRequest{
+		domain: nameservers,
 	}
 
-	d.SetId(domain)
-	return resourceSplitDNSNameserversRead(ctx, d, m)
+	if _, err := r.Client.DNS().UpdateSplitDNS(ctx, updateReq); err != nil {
+		diags.AddError("Failed to update DNS split nameservers", err.Error())
+		return
```
neinkeinkaffee, line 147: Did you put this return here intentionally (e.g. for symmetry/clarity)?

tailscale/resource_dns_split_nameservers.go
```diff
@@ -6,97 +6,165 @@ package tailscale
...
-	domain := d.Get("domain").(string)
+func (r *dnsSplitNameserversResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	var plan, state dnsSplitNameserversResourceData
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	plan.ID = state.ID
```
neinkeinkaffee, line 119: Something for discussion: I think instead of setting the plan ID explicitly in the `Update` here (and in other already migrated resources) we could just use the `UseStateForUnknown` plan modifier on the `id` field.

neinkeinkaffee (approved):
Left some nitpicks but looks good to me!

tailscale/resource_dns_split_nameservers.go
neinkeinkaffee: Actually ignore this! (It is better to keep it just in case rather than relying on the next person who puts statements after it remembering to put it back.)

tailscale/resource_dns_split_nameservers.go
alexwlchan: Good catch! No, that's just a temporary name I use while I have the old plugin SDK `schema` in the import list. I'll clean it up.

tailscale/resource_dns_split_nameservers.go
alexwlchan: Makes sense! Yeah, I think there's a few things we might want to go back and tidy up later to make everything consistent, once we've got more experience with the plugin framework and know what abstractions/helpers would be useful.

I'm going to leave this as-is for now but let's revisit once the migration is done.

PR #691, opened 2026-04-16
resource_device_authorization: preserve state ID on update

### LINKED ISSUE #690: `device_authorization`: Incorrect ID field in `Read`
When refreshing state from Tailscale, you can end up in a situation where `no manageable device matching this ID found (404)` for `tailscale_device_authorization`.  This is because it is using a stale `Id`. instead of the `DeviceId` argument during the Read.

https://github.com/tailscale/terraform-provider-tailscale/blob/64e90dbd05ec62de410925b2ef0728dcbeea16e0/tailscale/resource_device_authorization.go#L66-L76

Compare this to `tailscale_device_subnet_routes`, which is properly using the DeviceId:

https://github.com/tailscale/terraform-provider-tailscale/blob/64e90dbd05ec62de410925b2ef0728dcbeea16e0/tailscale/resource_device_subnet_routes.go#L84-L94

neinkeinkaffee:
This ensures that the state ID gets passed on correctly on Update. Also, previously, the resource didn't support un-authorising a device. We're removing that limitation because it dates from a time when this wasn't supported by the API. The two changes are bundled together because the latter makes adding a test for the former more straightforward.

Updates tailscale/corp#37231

This might also resolve #690. 

Terraform acceptance test: https://github.com/tailscale/corp/actions/runs/24505904516/job/71624246153

### DIFF
--- a/tailscale/resource_device_authorization.go
+++ b/tailscale/resource_device_authorization.go
@@ -48,6 +48,9 @@ func (d deviceAuthorizationResource) Schema(_ context.Context, _ resource.Schema
 		Attributes: map[string]schema.Attribute{
 			"id": schema.StringAttribute{
 				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
 			},
 			"device_id": schema.StringAttribute{
 				Required:    true,
@@ -106,13 +109,11 @@ func (d deviceAuthorizationResource) Create(ctx context.Context, req resource.Cr
 	deviceID := plan.DeviceID.ValueString()
 	authorized := plan.Authorized.ValueBool()
 
-	if authorized {
-		if err := d.Client.Devices().SetAuthorized(ctx, deviceID, true); err != nil {
-			resp.Diagnostics.AddError(
-				"Failed to authorize device",
-				"Could not authorize device with ID "+deviceID+": "+err.Error(),
-			)
-		}
+	if err := d.Client.Devices().SetAuthorized(ctx, deviceID, authorized); err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to authorize device",
+			"Could not authorize device with ID "+deviceID+": "+err.Error(),
+		)
 	}
 
 	plan.ID = types.StringValue(deviceID)
@@ -130,30 +131,15 @@ func (d deviceAuthorizationResource) Update(ctx context.Context, req resource.Up
 
 	deviceID := plan.DeviceID.ValueString()
 
-	device, err := d.Client.Devices().Get(ctx, deviceID)
-	if err != nil {
-		resp.Diagnostics.AddError(
-			"Failed to fetch device",
-			"Could not fetch device with device with ID"+deviceID+": "+err.Error(),
-		)
-		return
-	}
+	authorized := plan.Authorized.ValueBool()
 
-	// Currently, the Tailscale API only supports authorizing a device, but not un-authorizing one. So if the device
-	// data from the API states it is authorized then we can't do anything else here.
-	if device.Authorized {
-		plan.Authorized = types.BoolValue(true)
-		return
-	}
-
-	if err = d.Client.Devices().SetAuthorized(ctx, deviceID, true); err != nil {
+	if err := d.Client.Devices().SetAuthorized(ctx, deviceID, authorized); err != nil {
 		resp.Diagnostics.AddError(
 			"Failed to authorize device",
 			"Could not authorize device with ID"+deviceID+": "+err.Error(),
 		)
 		return
 	}
-	plan.Authorized = types.BoolValue(true)
 
 	diags = resp.State.Set(ctx, plan)
 	resp.Diagnostics.Append(diags...)
@@ -162,4 +148,15 @@ func (d deviceAuthorizationResource) Update(ctx context.Context, req resource.Up
 func (d deviceAuthorizationResource) Delete(_ context.Context, _ resource.DeleteRequest, _ *resource.DeleteResponse) {
 	// Since authorization cannot be removed at this point, deleting the resource will do nothing.
 	return
+}
+
+func (r deviceAuthorizationResource) ModifyPlan(_ context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
+	// If the entire plan is null, the resource is planned for destruction.
+	if req.Plan.Raw.IsNull() {
+		resp.Diagnostics.AddWarning(
+			"Resource Destruction Considerations",
+			"Applying this resource destruction will only remove the resource from the Terraform state and "+
+				"will not modify the device's authorization. ",
+		)
+	}
 }
--- a/tailscale/resource_device_authorization_test.go
+++ b/tailscale/resource_device_authorization_test.go
@@ -18,7 +18,7 @@ import (
 func TestAccTailscaleDeviceAuthorization(t *testing.T) {
 	const resourceName = "tailscale_device_authorization.test_authorization"
 
-	const testDeviceAuthorization = `
+	const testDeviceAuthorizationCreate = `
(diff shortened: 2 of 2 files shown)


tailscale/resource_device_authorization.go
```diff
@@ -48,6 +48,9 @@ func (d deviceAuthorizationResource) Schema(_ context.Context, _ resource.Schema
 		Attributes: map[string]schema.Attribute{
 			"id": schema.StringAttribute{
 				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
```
neinkeinkaffee, line 53: Need to ensure that the ID gets propagated on update. Going for the variant with the plan modifier here instead of "by hand" - as discussed elsewhere, we'll likely want to decide for one of the ways before the next release.

tailscale/resource_device_authorization.go
```diff
@@ -130,30 +131,15 @@ func (d deviceAuthorizationResource) Update(ctx context.Context, req resource.Up
...
-	// Currently, the Tailscale API only supports authorizing a device, but not un-authorizing one. So if the device
-	// data from the API states it is authorized then we can't do anything else here.
-	if device.Authorized {
-		plan.Authorized = types.BoolValue(true)
-		return
-	}
+	authorized := plan.Authorized.ValueBool()
 
-	if err = d.Client.Devices().SetAuthorized(ctx, deviceID, true); err != nil {
+	if err := d.Client.Devices().SetAuthorized(ctx, deviceID, authorized); err != nil {
```
neinkeinkaffee, line 136: Previously, the resource would check whether the device was already authorised and if so return immediately. Otherwise, it would call the API to authorise the device. The reason for this was that when the resource was introduced, the API didn't support un-authorising resources. Given that it does now, we can just make the call and update the state as we normally do.

tailscale/resource_device_authorization_test.go
```diff
@@ -28,18 +28,39 @@ func TestAccTailscaleDeviceAuthorization(t *testing.T) {
...
-			return fmt.Errorf("device with id %q is not authorized", rs.Primary.ID)
+	const testDeviceAuthorizationNextUpdate = `
+		data "tailscale_device" "test_device" {
+			name = "%s"
 		}
+		
+		resource "tailscale_device_authorization" "test_authorization" {
+			device_id = data.tailscale_device.test_device.id
+			authorized = true
+		}`
```
neinkeinkaffee, line 49: I've introduced a second update into the test that resets `authorized` to true so that we can ensure that the device doesn't get un-authorized on destroy. 

This actually raises the question whether we want to keep things that way or whether destroying the Terraform resource should result in un-authorizing the device. I think it shouldn't. But it's important to note in this context is that destroying the device_key resource _does_ set `key_expiry_disable` to false.

logan-hcg:
@neinkeinkaffee unfortunately this won't resolve #690, because that happens during refresh.  The issue is that if you replace the resource (ie the `device_id`) that the authorization should be working on, the state.Id is still caching the old (incorrect) id.  Example:

```
terraform state show tailscale_device_authorization.authorized

resource "tailscale_device_authorization" "authorized" {                                                                                                                                                                                                                                                                     
    authorized = true                                                                                                                                                                                                                                                                                                        
    device_id  = "nWXxdYuShi11CNTRL"                                                                                                                                                                                                                                                                                         
    id         = "ngu2nXQ4AE11CNTRL"                                                                                                                                                                                                                                                                                         
} 
```

neinkeinkaffee:
Thanks @logan-hcg now I understand the issue. That issue should be fixed with the next release because we have changed the provider to [recreate the resource](https://github.com/tailscale/terraform-provider-tailscale/blob/64e90dbd05ec62de410925b2ef0728dcbeea16e0/tailscale/resource_device_authorization.go#L55-L57) when the `device_id` changes.

logan-hcg:
> Thanks @logan-hcg now I understand the issue. That issue should be fixed with the next release because we have changed the provider to [recreate the resource](https://github.com/tailscale/terraform-provider-tailscale/blob/64e90dbd05ec62de410925b2ef0728dcbeea16e0/tailscale/resource_device_authorization.go#L55-L57) when the `device_id` changes.

@neinkeinkaffee thats great news, thank you!

tailscale/resource_device_authorization.go
mpminardi: I think that using `stringplanmodifier.UseStateForUnknown()` for ID values across the board would be a sane default and would lean more towards that.

In the [recommendations section for resource updates](https://developer.hashicorp.com/terraform/plugin/framework/resources/update#recommendations) they mention using this modifier for:

> Computed attributes that are known to not change during resource updates

which our ID values should fall under (and they show using this modifier for ID values elsewhere on the page).

tailscale/resource_device_authorization_test.go
mpminardi: Yeah this is a tricky one with how these resources / APIs are currently structured (namely, since this resource doesn't map back to an actual "object" but instead a setting on an object). Leaving the state as-is keeps the current behaviour of deletion, and I think makes more sense in this case.

One of the benefits of us migrating to the framework is that we can give [destroy plan diagnostics](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#resource-destroy-plan-diagnostics) for odd cases like this where a "destroy" is unintuitive in what it entails.

E.g., we could have something like:

```go
func (r deviceAuthorizationResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
	// If the entire plan is null, the resource is planned for destruction.
	if req.Plan.Raw.IsNull() {
		resp.Diagnostics.AddWarning(
			"Resource Destruction Considerations",
			"Applying this resource destruction will only remove the resource from the Terraform state "+
				"and will not modify the device's authorization. 
		)
	}
}
```

or something similar to give users an idea of what this entails (provided they use terraform 1.3 or later).

neinkeinkaffee: Good idea, I'll add that call-out on destroy!

neinkeinkaffee:
Another round of testing reveals that the RequiresReplace on the ID field doesn't actually resolve the issue described in #690 fully. When a device has gone away for good from the tailnet, we'll still have old ID and DeviceID in the state and a plan will fail on Read with `Could not read device authorization for device with ID nFOO123456DEVEL: no manageable device matching this ID found (404)`. resource_device_subnet_routes solves this by [removing the resource from state](https://github.com/tailscale/terraform-provider-tailscale/blob/e62cf6fbe62d4f15bae69ccf733ce83d0ba7575d/tailscale/resource_device_subnet_routes.go#L97-L101) in order to repopulate it subsequently, which may be the pattern to emulate for the other device resources. In any case, I'll address that in a separate PR.

mpminardi:
> Another round of testing reveals that the RequiresReplace on the ID field doesn't actually resolve the issue described in #690 fully. When a device has gone away for good from the tailnet, we'll still have old ID and DeviceID in the state and a plan will fail on Read with `Could not read device authorization for device with ID nFOO123456DEVEL: no manageable device matching this ID found (404)`. resource_device_subnet_routes solves this by [removing the resource from state](https://github.com/tailscale/terraform-provider-tailscale/blob/e62cf6fbe62d4f15bae69ccf733ce83d0ba7575d/tailscale/resource_device_subnet_routes.go#L97-L101) in order to repopulate it subsequently, which may be the pattern to emulate for the other device resources. In any case, I'll address that in a separate PR.

Good catch! This is indeed something we want more broadly in the provider once we audit on an endpoint-by-endpoint basis that we are returning a 404 when expected (more of a sanity measure since many of the endpoints are fairly old and haven't been deeply analyzed by current folks on the team). 

See https://github.com/tailscale/corp/issues/26072 as well.

mpminardi (approved):

PR #693, opened 2026-04-20
resource_webhook: migrate to plugin framework

alexwlchan:
Fixes tailscale/corp#37247

I haven't been able to test it yet because I wrote it on an airplane, but it at least compiles. 😅

### DIFF
--- a/tailscale/resource_webhook.go
+++ b/tailscale/resource_webhook.go
@@ -6,54 +6,80 @@ package tailscale
 import (
 	"context"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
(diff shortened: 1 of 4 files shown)


tailscale/resource_webhook.go
```diff
@@ -6,54 +6,84 @@ package tailscale
...
 				Description: "The provider type of the endpoint URL. This determines the payload format sent to the destination. Valid values are `slack`, `mattermost`, `googlechat`, and `discord`.",
 				Optional:    true,
-				ForceNew:    true,
-				ValidateFunc: validation.StringInSlice(
-					[]string{
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				// TODO(alexc): The plugin SDK validator set `ignoreCase: false`;
+				// check that's still the case with the plugin framework.
```
mpminardi: `stringvalidator.OneOf` does indeed seem to be what we want here! It looks like there is also a case insensitive version if we ever need it (see https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator#OneOfCaseInsensitive).

mpminardi (approved):

PR #697, opened 2026-04-27
tailscale: auto-recover from device not found

neinkeinkaffee:
Remove a resource from state on Read if the backing device cannot be found. This enables recreation without manual intervention by the user (through `terraform state rm` or similar). We already do this for tailscale_device_subnet_routes, tailscale_service and tailscale_tailnet_key. 

Updates #690

Terraform acceptance test: https://github.com/tailscale/corp/actions/runs/24981486773/job/73144508080

### DIFF
--- a/tailscale/resource_device_authorization.go
+++ b/tailscale/resource_device_authorization.go
@@ -12,11 +12,13 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"tailscale.com/client/tailscale/v2"
 )
 
 var (
 	_ resource.Resource                = &deviceAuthorizationResource{}
 	_ resource.ResourceWithImportState = &deviceAuthorizationResource{}
+	_ resource.ResourceWithModifyPlan  = &deviceAuthorizationResource{}
 )
 
 type deviceAuthorizationResourceModel struct {
@@ -78,6 +80,12 @@ func (d deviceAuthorizationResource) Read(ctx context.Context, req resource.Read
 
 	device, err := d.Client.Devices().Get(ctx, deviceID)
 	if err != nil {
+		// If the device is not found, remove from the state so we can create it again.
+		if tailscale.IsNotFound(err) {
+			resp.State.RemoveResource(ctx)
+			return
+		}
+
 		resp.Diagnostics.AddError(
 			"Failed to fetch device authorization",
 			"Failed to fetch authorization for device with with ID "+deviceID+": "+err.Error(),
@@ -153,12 +161,14 @@ func (d deviceAuthorizationResource) Delete(_ context.Context, _ resource.Delete
 	return
 }
 
-func (r deviceAuthorizationResource) ModifyPlan(_ context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
+func (d deviceAuthorizationResource) ModifyPlan(_ context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
+	// If the entire plan is null, the resource is planned for destruction.
 	if req.Plan.Raw.IsNull() {
 		resp.Diagnostics.AddWarning(
 			"Resource Destruction Considerations",
 			"Applying this resource destruction will only remove the resource from the Terraform state and "+
 				"will not modify the device's authorization. ",
 		)
+		return
 	}
 }
--- a/tailscale/resource_device_key.go
+++ b/tailscale/resource_device_key.go
@@ -116,6 +116,12 @@ func (d deviceKeyResource) Read(ctx context.Context, req resource.ReadRequest, r
 
 	device, err := d.Client.Devices().Get(ctx, deviceID)
 	if err != nil {
+		// If the device is not found, remove from the state so we can create it again.
+		if tailscale.IsNotFound(err) {
+			resp.State.RemoveResource(ctx)
+			return
+		}
+
 		resp.Diagnostics.AddError(
 			"Failed to fetch device key",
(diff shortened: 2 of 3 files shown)


tailscale/resource_device_key.go
```diff
@@ -167,3 +178,28 @@ func (d deviceKeyResource) Update(ctx context.Context, req resource.UpdateReques
...
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	deviceID := plan.DeviceID.ValueString()
+	if _, err := d.Client.Devices().Get(ctx, deviceID); err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to fetch devices",
```
alexwlchan:
```suggestion
			"Failed to fetch device key",
```

tailscale/resource_device_tags.go
```diff
@@ -183,7 +191,32 @@ func (d deviceTagsResource) Delete(ctx context.Context, req resource.DeleteReque
...
+	diags := req.Plan.Get(ctx, &plan)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	deviceID := plan.DeviceID.ValueString()
+	if _, err := d.Client.Devices().Get(ctx, deviceID); err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to fetch device",
```
alexwlchan:
```suggestion
			"Failed to fetch device tags",
```

tailscale/resource_device_authorization.go
```diff
@@ -78,6 +80,12 @@ func (d deviceAuthorizationResource) Read(ctx context.Context, req resource.Read
 
 	device, err := d.Client.Devices().Get(ctx, deviceID)
 	if err != nil {
+		// If the device is not found, remove from the state so we can create it again.
+		if tailscale.IsNotFound(err) {
```
alexwlchan, line 84: nit: I would move this `if` statement out of `if err != nil` to reduce the nesting and create two distinct branches, but it's a style choice, not required.

alexwlchan (commented):
Can we write a test for this behaviour?

tailscale/resource_device_key.go
neinkeinkaffee: Worth consolidating this for all devices: I've created an issue: https://github.com/tailscale/corp/issues/40809

neinkeinkaffee:
> Can we write a test for this behaviour?

We could potentially fudge the scenarios that exercise this but would need to come up with a sensible way of making the `testServer` do something more sophisticated than just cycle through canned responses.

alexwlchan:
> would need to come up with a sensible way of making the testServer do something more sophisticated than just cycle through canned responses

Maybe an acceptance test where we use the `PreConfig` to delete the device outside Terraform? I'm not sure if we have easy access to an API client for the acceptance test server.

neinkeinkaffee:
> Maybe an acceptance test where we use the PreConfig to delete the device outside Terraform?

Certainly possible. I had in the back of my mind that we've so far avoided deleting machines in order to not have to backfill them between acceptance test runs, so that'd have to be taken care of...

tailscale/resource_device_authorization.go
neinkeinkaffee: I don't have a strong preference for either but we do use this pattern in three other places in addition to the three that this PR would be adding (resource_device_subnet_routes.go, resource_service.go and resource_tailnet_key.go). If we do decide to change it I'd prefer to do it in a one wholesale pass over all resources concerned in a separate PR.

tailscale/resource_device_authorization.go
```diff
@@ -114,6 +122,7 @@ func (d deviceAuthorizationResource) Create(ctx context.Context, req resource.Cr
 			"Failed to authorize device",
 			"Could not authorize device with ID "+deviceID+": "+err.Error(),
 		)
+		return
```
neinkeinkaffee, line 125: This missing return on error was a drive-by but I've since raised a separate PR for it: https://github.com/tailscale/terraform-provider-tailscale/pull/710

neinkeinkaffee:
I've come to the conclusion that sticking to just removing the resource from state on Read (which is what tailscale_device_subnet_routes, tailscale_service and tailscale_tailnet_key are already doing) is good enough for tailscale_device_authorization, tailscale_device_key, tailscale_device_tags. So I've removed the overcautious pre-flight double-checking whether the updated device ID exists. The main reason is my realisation that after all, we only commit the removal from state if the user _applies_ the plan, not if they just plan and then realise that something is wrong. This should be enough protection against accidentally passing the wrong tailnet's API key (or hypothetical transient 404s from the control plane), as it should be the responsibility of the user to verify the plan.

neinkeinkaffee:
@alexwlchan I've hit "resolved" on your comments with suggested edits for the error messages because the specific errors they were relating to are gone with the removal of the ModifyPlan step, and because the general issue of coming up with a consistent error message pattern for devices will be addressed separately.

Edit: and here's the PR to do that (consolidate the error messages): https://github.com/tailscale/terraform-provider-tailscale/pull/713

tailscale/resource_device_authorization.go
```diff
@@ -153,12 +161,14 @@ func (d deviceAuthorizationResource) Delete(_ context.Context, _ resource.Delete
 	return
 }
 
-func (r deviceAuthorizationResource) ModifyPlan(_ context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
+func (d deviceAuthorizationResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
```
mpminardi: Nit: I think we want to leave the `ctx` as `_` here since it is unused.

mpminardi (approved):

tailscale/resource_device_authorization.go
neinkeinkaffee: Blame the most recent unskilfully performed rebase...

PR #699, opened 2026-04-27
resource_contacts: migrate to plugin framework

alexwlchan:
Fixes tailscale/corp#37230


tailscale/resource_contacts.go
```diff
@@ -6,179 +6,211 @@ package tailscale
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
 		},
-		Schema: map[string]*schema.Schema{
-			"account": {
-				Type:        schema.TypeSet,
+		Blocks: map[string]schema.Block{
+			"account": schema.ListNestedBlock{
```
mpminardi: We'll need these to be `SetNestedBlock` and do a `List` -> `Set` throughout to not break backwards compatibility with the existing shape of this resource (see https://developer.hashicorp.com/terraform/plugin/best-practices/versioning#example-major-number-increments).

alexwlchan: Fixed!

mpminardi (approved):

PR #700, opened 2026-04-27
data_source_{user,users}: migrate to plugin framework

alexwlchan:
Fixes tailscale/corp#37226
Fixes tailscale/corp#37227

### DIFF
--- a/tailscale/data_source_user.go
+++ b/tailscale/data_source_user.go
@@ -5,141 +5,96 @@ package tailscale
 
 import (
 	"context"
-	"time"
+	"maps"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 
 	"tailscale.com/client/tailscale/v2"
 )
 
-var commonUserSchema = map[string]*schema.Schema{
-	"display_name": {
-		Type:        schema.TypeString,
-		Description: "The name of the user.",
-		Computed:    true,
-	},
-	"profile_pic_url": {
-		Type:        schema.TypeString,
-		Description: "The profile pic URL for the user.",
-		Computed:    true,
-	},
-	"tailnet_id": {
-		Type:        schema.TypeString,
-		Description: "The tailnet that owns the user.",
-		Computed:    true,
-	},
-	"created": {
-		Type:        schema.TypeString,
-		Description: "The time the user joined their tailnet.",
-		Computed:    true,
-	},
-	"type": {
-		Type:        schema.TypeString,
-		Description: "The type of relation this user has to the tailnet associated with the request.",
-		Computed:    true,
-	},
-	"role": {
-		Type:        schema.TypeString,
-		Description: "The role of the user.",
-		Computed:    true,
-	},
-	"status": {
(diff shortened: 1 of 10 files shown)


tailscale/data_source_user.go
```diff
@@ -5,141 +5,96 @@ package tailscale
...
+		if user == nil {
+			resp.Diagnostics.AddError("User not found", "Could not find user with login name: "+data.LoginName.ValueString())
+			return
+		}
+	} else {
+		resp.Diagnostics.AddError(
+			"please specify an id or login_name for the user",
+			"please specify an id or login_name for the user",
+		)
 	}
```
neinkeinkaffee, line 96: This is a `// we should never get here` case because of the ExactlyOneOf validator, right? But probably better safe than sorry. Now I wonder whether data_source_device should get a safeguard like this too...

neinkeinkaffee: ("This" referring to the last else)

neinkeinkaffee:
Is the omission of tests that check for drift on upgrade to the plugin framework intentional?

alexwlchan:
> Is the omission of tests that check for drift on upgrade to the plugin framework intentional?

No, it's a mistake, I'll add them.

tailscale/data_source_user.go
alexwlchan: Good point! I'll replace it with an unreachable panic, and add a couple of tests for invalid configurations.

alexwlchan:
@neinkeinkaffee I've addressed your feedback, please take another look. :)

tailscale/data_source_users_test.go
```diff
@@ -119,4 +122,29 @@ func TestAccTailscaleUsers(t *testing.T) {
...
+
+	// Check that the data sources behave the same between the plugin SDK
+	// and the plugin framework.
+	checkDataSourceIsUnchangedInPluginFramework(t, `data "tailscale_users" "all_users" {}`)
+	checkDataSourceIsUnchangedInPluginFramework(t, userDataSources.String())
+}
+
+// userToMap converts the given user into a map representing the user as a
+// resource in Terraform. This omits the "id" which is expected to be set
+// using [schema.ResourceData.SetId].
```
neinkeinkaffee, line 134:
```suggestion
// resource in Terraform.
```

tailscale/model_user.go
```diff
@@ -0,0 +1,85 @@
...
+	Created            types.String `tfsdk:"created"`
+	Type               types.String `tfsdk:"type"`
+	Role               types.String `tfsdk:"role"`
+	Status             types.String `tfsdk:"status"`
+	DeviceCount        types.Int32  `tfsdk:"device_count"`
+	LastSeen           types.String `tfsdk:"last_seen"`
+	CurrentlyConnected types.Bool   `tfsdk:"currently_connected"`
+}
+
+func (d *userDataSourceModel) setUserProperties(user *tailscale.User) {
```
neinkeinkaffee: Just curious: any reason for departing from the `toXYZDataSourceModel`-pattern in the other places here? The one difference I can see is that we don't need to catch and return error diagnostics?

tailscale/data_source_users.go
```diff
@@ -5,99 +5,125 @@ package tailscale
...
+	data.Users = make([]userDataSourceModel, 0, len(users))
+	for _, u := range users {
+		userData := userDataSourceModel{}
+		userData.setUserProperties(&u)
+		data.Users = append(data.Users, userData)
 	}
 
-	d.SetId(createUUID())
-	return nil
+	data.ID = types.StringValue(createUUID())
```
neinkeinkaffee, line 126: Same thought as for the devices data source that we might want to look into creating a stable ID somehow  instead of setting a UUID that changes on every Read. I'll create a separate issue to follow up!

neinkeinkaffee (approved):
Looks good from my point of view now (except we're missing that copyright header in data_souce_user_test.go).

tailscale/model_user.go
alexwlchan: Just an oversight; I'll change it.

PR #701, opened 2026-04-27
data_source_devices: migrate to plugin framework

alexwlchan:
I copied some of the ideas from #700 and the `data_source_{user, users}` resources – rather than duplicating the schema in both files, we can reuse it and get the benefit of improved documentation to boot!

Fixes tailscale/corp#37225

### DIFF
--- a/tailscale/data_source_device.go
+++ b/tailscale/data_source_device.go
@@ -7,170 +7,74 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
-	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"tailscale.com/client/tailscale/v2"
-	"tailscale.com/types/bools"
 )
 
-type deviceDataSourceModel struct {
-	Hostname types.String `tfsdk:"hostname"`
-	Name     types.String `tfsdk:"name"`
-	WaitFor  types.String `tfsdk:"wait_for"`
-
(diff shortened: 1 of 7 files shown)


tailscale/data_source_device.go
```diff
@@ -251,100 +155,16 @@ func (d deviceDataSource) Read(ctx context.Context, req datasource.ReadRequest,
...
-// using [schema.ResourceData.SetId].
-func deviceToMap(device *tailscale.Device) map[string]any {
-	var lastSeen string
-	if device.LastSeen == nil {
-		lastSeen = ""
-	} else {
-		lastSeen = device.LastSeen.Format(time.RFC3339)
-	}
+	device.deviceDataSourceModel = apiData
+	device.ID = apiData.ID
```
neinkeinkaffee: This assignment seems redundant?

neinkeinkaffee: Doesn't the line before already set the ID?

neinkeinkaffee:
> rather than duplicating the schema in both files, we can reuse it and get the benefit of improved documentation to boot

Nice!

tailscale/data_source_device.go
alexwlchan: Yup, I started writing out the fields manually and then backtracked, but this one got left behind. Fixing now.

tailscale/data_source_devices.go
```diff
@@ -5,212 +5,140 @@ package tailscale
...
+		}
 
-	if err = d.Set("devices", deviceMaps); err != nil {
-		return diag.FromErr(err)
+		data.Devices = append(data.Devices, deviceModel)
 	}
 
-	d.SetId(createUUID())
-	return nil
+	data.ID = types.StringValue(createUUID())
```
neinkeinkaffee, line 142: Just realised that this resets the ID in state and therefore creates a diff on each Read... I'll create a separate issue to look into changing that! All I can find in the official docs is the HashiCups provider refraining from setting the ID for this kind of [list/query data sources](https://developer.hashicorp.com/terraform/tutorials/community-providers/providers-plugin-framework-lab#add-the-read-function) at all but I'm not sure that's the answer.

neinkeinkaffee (approved):

tailscale/data_source_devices.go
neinkeinkaffee: https://github.com/tailscale/terraform-provider-tailscale/issues/714

PR #707, opened 2026-04-28
resource_posture_integration: migrate to plugin framework

zofrex:
Migrate the posture_integration resource to the plugin framework.

Test plan: run all tests with acceptance testing enabled, e.g. `make testacc`

Updates tailscale/corp#37243

PR #709, opened 2026-04-29
resource_webhook: handle null provider

zofrex:
(Stacked on #707 so I can re-use `StringValueNullIfEmpty` from there)

I added a test to check that not supplying a provider works, but it failed because an empty string comes back from the client, and the state has a null instead.

Updated the provider to coerce an empty string coming back from the client to a null value.

Test plan: run all acceptance tests. Alternatively, create a terraform file that creates a webhook without setting the `provider` field and ensure that it does not still show a diff after a plan.

### DIFF
--- a/tailscale/resource_webhook.go
+++ b/tailscale/resource_webhook.go
@@ -191,7 +191,7 @@ func (r *webhookResource) Read(ctx context.Context, req resource.ReadRequest, re
 	}
 
 	state.EndpointURL = types.StringValue(webhook.EndpointURL)
-	state.ProviderType = types.StringValue(string(webhook.ProviderType))
+	state.ProviderType = StringValueNullIfEmpty(string(webhook.ProviderType))
 
 	subscriptions, diags := types.SetValueFrom(ctx, types.StringType, webhook.Subscriptions)
 	resp.Diagnostics.Append(diags...)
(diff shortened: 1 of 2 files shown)


tailscale/resource_webhook.go
```diff
@@ -191,7 +191,7 @@ func (r *webhookResource) Read(ctx context.Context, req resource.ReadRequest, re
 	}
 
 	state.EndpointURL = types.StringValue(webhook.EndpointURL)
-	state.ProviderType = types.StringValue(string(webhook.ProviderType))
+	state.ProviderType = StringValueNullIfEmpty(string(webhook.ProviderType))
```
zofrex, line 194: This doesn't have the problem of creating a spurious diff when set to the empty string, because this value is not permitted to be empty. It must be null, or one of the permitted values.

alexwlchan:
I've rebased against main and force-pushed so I can review this properly.

alexwlchan (approved):

zofrex:
Why doesn't this have an issue with the empty string `""` being in previous state created by the SDK?

Well from what I can tell, if you have such previous state, and do an apply with the new provider, and it sets the value to `null`, the state is silently updated without a diff from Terraform, I guess to smooth over the transition from empty values to nulls in the state file.

In some of the other resources, it's valid to actually set a value to `""` and then we have to care about what's in the state file to not create spurious diffs, but in this resource, it has to have a value, so we can ignore that case: if the API says `""`, we always store `nil`, and we don't have to worry about that case.

PR #715, opened 2026-05-05
data_source_service: migrate to plugin framework

alexwlchan:
Fixes tailscale/corp#40811

### DIFF
--- a/tailscale/data_source_service.go
+++ b/tailscale/data_source_service.go
@@ -6,78 +6,101 @@ package tailscale
 import (
 	"context"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-
-	"tailscale.com/client/tailscale/v2"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-func dataSourceService() *schema.Resource {
-	return &schema.Resource{
+// NewServiceDataSource() returns a new Services data source.
+func NewServiceDataSource() datasource.DataSource {
+	return &dataSourceService{}
+}
+
+type dataSourceService struct {
+	DataSourceBase
+}
+
+// Metadata defines the data source name as it appears in Terraform configurations.
+func (d dataSourceService) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_service"
+}
+
+// Schema defines a schema describing what data is available in the data source response.
+func (d dataSourceService) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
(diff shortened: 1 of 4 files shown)


tailscale/data_source_service.go
```diff
@@ -6,78 +6,101 @@ package tailscale
...
 				Description: "The ports that the Service listens on.",
 				Computed:    true,
-				Elem: &schema.Schema{
-					Type: schema.TypeString,
-				},
+				ElementType: types.StringType,
 			},
-			"tags": {
-				Type:        schema.TypeSet,
+			"tags": schema.SetAttribute{
```
alexwlchan, line 55: The Go SDK returns addresses/ports/tags all as slices; it's not obvious to me that this should be a set rather than a list – but I'm going to leave it as-is for this PR.

tailscale/data_source_service_test.go
```diff
@@ -101,4 +102,20 @@ func TestAccTailscaleDataSourceService(t *testing.T) {
 			},
 		},
 	})
+
+	// Note: we don't need to test that this data source is unchanged in the
+	// plugin framework, because it was not introduced until after we'd moved
+	// from the plugin SDK to the plugin framework.
```
alexwlchan, line 108: I was extremely confused when I tried to add the test and got errors about `tailscale_service`  not being defined, until I realised those errors were coming from the old version of the provider. 😂

tailscale/data_source_service.go
zofrex: Set seems right conceptually for tags for sure. For addresses, list seems right in this case, because positions 1 & 2 in the address have special meaning (alternatively we could expose them separately as ipv4 and ipv6 rather than an array of both).

My read of ports is that order is unimportant as well, so maybe that should also be a set?

tailscale/data_source_service.go
```diff
@@ -6,78 +6,101 @@ package tailscale
...
-		"tags":    svc.Tags,
-	})
+	data.ID = types.StringValue(svc.Name)
+	data.Name = types.StringValue(svc.Name)
+
+	addrs, diags := types.ListValueFrom(ctx, types.StringType, svc.Addrs)
+	resp.Diagnostics.Append(diags...)
+	data.Addrs = addrs
+
+	data.Comment = types.StringValue(svc.Comment)
```
zofrex: It's optional so I think it would be good to coalesce empty string to nil here (less important as a data source than a resource, but good to be consistent about it)

tailscale/data_source_service.go
```diff
@@ -6,78 +6,101 @@ package tailscale
...
+
+	ports, diags := types.ListValueFrom(ctx, types.StringType, svc.Ports)
+	resp.Diagnostics.Append(diags...)
+	data.Ports = ports
+
+	tags, diags := types.SetValueFrom(ctx, types.StringType, svc.Tags)
```
zofrex, line 101: If there are no tags does this give us nil or the empty set?

zofrex: (From testing: empty set)

tailscale/data_source_service.go
alexwlchan: I think the commit I just pushed fixes this? Let me know if not

zofrex: (After chatting with Mario about it, this is fine)

alexwlchan: Okay, I'll yank that commit.

zofrex (approved):

alexwlchan:
I swear I merged this already. 🤔

PR #716, opened 2026-05-05
resource_oauth_client: migrate to plugin framework

alexwlchan:
Fixes tailscale/corp#37242

### DIFF
--- a/tailscale/resource_oauth_client.go
+++ b/tailscale/resource_oauth_client.go
@@ -7,205 +7,199 @@ import (
 	"context"
 	"time"
 
-	"github.com/hashicorp/go-cty/cty"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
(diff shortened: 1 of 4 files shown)


tailscale/resource_oauth_client.go
```diff
@@ -7,205 +7,205 @@ import (
...
 
-	key, err := client.Keys().SetOAuthClient(ctx, d.Id(),
-		tailscale.SetOAuthClientRequest{
-			Description: description.(string),
-			Scopes:      scopes,
-			Tags:        tags,
-		})
+	var scopes, tags []string
+	resp.Diagnostics.Append(plan.Scopes.ElementsAs(ctx, &scopes, false)...)
+	resp.Diagnostics.Append(plan.Tags.ElementsAs(ctx, &tags, false)...)
```
neinkeinkaffee, line 173: Did you verify that empty tags don't result in a spurious diff and that we don't need to copy the plan modifier pattern that the posture integration uses for the string case?

neinkeinkaffee:
Looks good, just the one question as to whether this one doesn't need a plan modifier to avoid spurious diffs for empty tags (don't have the conditions under which it arises fully in the back of my head, whether the attribute optional was enough to trigger it or it also needed to be computed).

tailscale/resource_oauth_client.go
alexwlchan: I did not! I'll check and add a test for it.

alexwlchan: Yup, looks okay; I added a test for that case and the case where somebody omits `scopes`.

neinkeinkaffee:
I had been thinking of the drift between versions backed by SDK v2 and framework SDK, so I thought testing that would be a case for `checkResourceIsUnchangedInPluginFramework`. But anyways, looks like we're good and the framework doesn't want a null here. Fine by me!

neinkeinkaffee (approved):

alexwlchan:
> I had been thinking of the drift between versions backed by SDK v2 and framework SDK

Ah, I see what you mean! I've added a test for that too, but no changes required.

PR #717, opened 2026-05-05
Add unit tests for plan modifiers

alexwlchan:
For migrating the `tailscale_acl` resource, we'll need a HuJSON plan modifier that ignores changes to strings which produce equivalent HuJSON strings; this code is getting complicated enough that I would like some tests as a safety net.

### DIFF
--- a/tailscale/plan_modifiers.go
+++ b/tailscale/plan_modifiers.go
@@ -17,6 +17,9 @@ var (
 	_ planmodifier.String = PreserveEmptyStringAsNull{}
 )
 
+// jsonSemanticDiffModifier is a plan modifier that will treat strings as
+// equivalent if they correspond to the same JSON value, and differ only
+// in whitespace.
 type jsonSemanticDiffModifier struct{}
 
 func (m jsonSemanticDiffModifier) Description(_ context.Context) string {
--- a/tailscale/plan_modifiers_test.go
+++ b/tailscale/plan_modifiers_test.go
@@ -1,0 +1,139 @@
+// Copyright (c) David Bond, Tailscale Inc, & Contributors
+// SPDX-License-Identifier: MIT
+
+package tailscale
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func TestJsonSemanticDiffModifier(t *testing.T) {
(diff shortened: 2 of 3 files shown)


tailscale/plan_modifiers.go
mpminardi, line 74: I know that this was not modified in this PR, but having looked at this a bit closer I think we want something shaped more like:

```suggestion
	// Early return if the configuration value is unknown
	if req.ConfigValue.IsUnknown() {
		return
	}

	if req.StateValue.IsNull() && req.ConfigValue.IsNull() {
		// State is null and config is null, plan as null
		resp.PlanValue = types.StringNull()
	} else if req.StateValue.ValueString() == "" && req.ConfigValue.ValueString() == "" {
		// State is empty and config is empty, plan as empty
		resp.PlanValue = types.StringValue("")
	} else if req.StateValue.ValueString() == "" && req.ConfigValue.IsNull() {
		// State is empty and config is null, plan as empty (bad SDK -> framework upgrade case)
		resp.PlanValue = types.StringValue("")
	}
```

This is similar to the above (but more explicitly laid out) with the difference that we do not override `null` values to `""` (which was the case previously due to `req.StateValue.ValueString()` being `""` for both empty and `null` string values it seems).

tailscale/plan_modifiers_test.go
```diff
@@ -0,0 +1,139 @@
...
+			state:        types.StringValue(""),
+			config:       types.StringUnknown(),
+			expectedPlan: types.StringUnknown(),
+		},
+		{
+			name:         "empty-string-in-state-null-plan",
+			state:        types.StringValue(""),
+			config:       types.StringNull(),
+			expectedPlan: types.StringValue(""),
+		},
```
mpminardi, line 96: We'll want to add a case shaped like:

```go
		{
			name:         "null-string-in-state-null-plan",
			state:        types.StringNull(),
			config:       types.StringNull(),
			expectedPlan: types.StringNull(),
		},
```

to test the case mentioned in https://github.com/tailscale/terraform-provider-tailscale/pull/717/changes#r3192984325

tailscale/plan_modifiers.go
mpminardi: Ah scratch that, I forgot that we need the `null` -> `""` conversion specifically for handling some bad diff rendering on Terraform's end. Will think through this again some more in the morning 😅 In the meantime I think this is good to leave as-is.

mpminardi (approved):

PR #718, opened 2026-05-05
provider: auto-discover OIDC tokens from runtime for workload identity federation

keeleysam:

Operators using workload identity federation today still have to plumb the OIDC token through `identity_token` or `identity_token_environment_variable_name`. That works, but every consumer ends up reimplementing the "fetch OIDC token from the runtime" step that the Tailscale CLI / tsnet path already does for free via `tailscale.com/wif.ObtainProviderToken`.

This PR adds an `audience` provider argument. When `oauth_client_id` and `audience` are both set (and no static token is supplied), the provider's `IDTokenFunc` calls `wif.ObtainProviderToken(ctx, audience)`, which detects the runtime (GitHub Actions, AWS via EC2 IMDSv2 or ECS, GCP via metadata server) and returns a fresh OIDC token. The token is then exchanged at Tailscale's token-exchange endpoint the same way the existing static-token federated identity flow already works:

```hcl
provider "tailscale" {
  oauth_client_id = "my_client_id"
  audience        = "my_audience"
  tailnet         = "example.com"
}
```

Follows up on #567, which introduced the static-token federated identity flow.

Fixes #


danni-popova (approved):
LGTM! Thank you for adding this!

PR #720, opened 2026-05-06
all: extract validators; add retryValidator tests; add invalid data_source_device config tests

alexwlchan:
For https://github.com/tailscale/corp/issues/37228 I need to write a custom plan modifier and a custom validator, and I'll want to write some unit tests.

This extracts the existing custom validators into `validators.go` and adds a test harness; it also adds some unit tests for where they're used in `data_source_device`. Plus some minor tidies in the plan modifier tests.

### DIFF
--- a/tailscale/data_source_4via6.go
+++ b/tailscale/data_source_4via6.go
@@ -5,10 +5,8 @@ package tailscale
 
 import (
 	"context"
-	"net"
 	"net/netip"
 
-	"github.com/hashicorp/terraform-plugin-framework-validators/helpers/validatordiag"
 	"github.com/hashicorp/terraform-plugin-framework-validators/int32validator"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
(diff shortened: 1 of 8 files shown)


tailscale/data_source_4via6.go
```diff
@@ -102,30 +100,3 @@ func (d dataSource4via6) Read(ctx context.Context, req datasource.ReadRequest, r
 
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
 }
-
-// cidrValidator is a [validator.String] for CIDR addresses.
-type cidrValidator struct{}
```
mpminardi, line 107: As an aside: did we investigate whether using the CIDR custom type from https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-nettypes@v0.3.0/cidrtypes would work here? I think it has ~similar validation built in.

mpminardi (approved):

tailscale/data_source_4via6.go
alexwlchan: I did not, because I didn't know it existed. It's worth a look, but that can be a follow-up PR.

PR #726, opened 2026-05-06
resource_tailnet_key: migrate to plugin framework

zofrex:
Updates tailscale/corp#37245

Test plan: run all tests with acceptance testing enabled.


tailscale/resource_tailnet_key.go
```diff
@@ -5,182 +5,197 @@ package tailscale
...
+		resp.Diagnostics.AddError("Failed to create key", fmt.Sprintf("Error creating tailnet key: %s", err.Error()))
+		return
 	}
 
-	d.SetId(key.ID)
+	plan.ID = types.StringValue(key.ID)
+	plan.Key = types.StringValue(key.Key)
+	plan.CreatedAt = types.StringValue(key.Created.Format(time.RFC3339))
+	plan.ExpiresAt = types.StringValue(key.Expires.Format(time.RFC3339))
+	plan.Expiry = types.Int64Value(int64(*key.ExpirySeconds))
```
mpminardi: I think we'll only want to set this if `key.ExpirySeconds` is non-nil:


```suggestion
    if key.ExpirySeconds != nil {
	     plan.Expiry = types.Int64Value(int64(*key.ExpirySeconds))
	}
```

tailscale/resource_tailnet_key.go
```diff
@@ -199,103 +214,120 @@ func shouldRecreateIfInvalid(reusable bool, recreateIfInvalid string) bool {
...
+	state.Ephemeral = types.BoolValue(key.Capabilities.Devices.Create.Ephemeral)
+	tags, diags := types.SetValueFrom(ctx, types.StringType, key.Capabilities.Devices.Create.Tags)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
 	}
+	state.Tags = tags
+	state.Preauthorized = types.BoolValue(key.Capabilities.Devices.Create.Preauthorized)
+	state.Key = types.StringValue(key.Key)
+	state.Expiry = types.Int64Value(int64(*key.ExpirySeconds))
```
mpminardi: Same note as above RE: only setting if `ExpirySeconds` is non-nil.

tailscale/resource_tailnet_key.go
zofrex: Can skip the 'if' with a pointer value:

```
types.Int64PointerValue((*int64)(key.ExpirySeconds))
```

(Also this will always be non-nil for an auth key — updated the test code to return a value here in the mock server, as it should. But still updated the code so it won't _crash_ if the API replies with a bad value here)

tailscale/resource_tailnet_key.go
```diff
@@ -5,182 +5,212 @@ package tailscale
...
+	plan.ExpiresAt = types.StringValue(key.Expires.Format(time.RFC3339))
+	plan.Expiry = types.Int64PointerValue((*int64)(key.ExpirySeconds))
+	plan.Invalid = types.BoolValue(key.Invalid)
+	plan.UserID = types.StringValue(key.UserID)
 
-	if err = d.Set("expires_at", key.Expires.Format(time.RFC3339)); err != nil {
-		return diagnosticsError(err, "Failed to set expires_at")
-	}
+	diags = resp.State.Set(ctx, plan)
+	resp.Diagnostics.Append(diags...)
```
mpminardi: Nit:

```suggestion
	resp.Diagnostics.Append(resp.State.Set(ctx, plan)...)
```

mpminardi (approved):

PR #732, opened 2026-05-12
tailscale: remove PreserveEmptyStringAsNull plan modifier

mpminardi:
Remove PreserveEmptyStringAsNull plan modifier and instead use default values for the empty value of the given type. This is essentially equivalent for resources that are migrating off of the plugin SDKv2 and is the more common pattern that we have consolidated on throughout the rest of the migration effort.

Updates https://github.com/tailscale/corp/issues/37032


mpminardi:
Acceptance tests: https://github.com/tailscale/corp/actions/runs/25710557617

zofrex (approved):

PR #737, opened 2026-05-12
provider_sdk: migrate validateProviderCreds to use Go error types

alexwlchan:
This decouples it from the plugin SDK.

Updates tailscale/corp#37221

### DIFF
--- a/tailscale/provider_sdk.go
+++ b/tailscale/provider_sdk.go
@@ -139,8 +139,8 @@ func providerConfigure(_ context.Context, provider *schema.Provider, d *schema.R
 	}
 	audience := d.Get("audience").(string)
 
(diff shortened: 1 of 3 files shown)


tailscale/provider_sdk.go
```diff
@@ -163,17 +163,17 @@ func providerConfigure(_ context.Context, provider *schema.Provider, d *schema.R
 	return &client, nil
 }
 
-func validateProviderCreds(apiKey, oauthClientID, oauthClientSecret, idToken, audience string) diag.Diagnostics {
+func validateProviderCreds(apiKey, oauthClientID, oauthClientSecret, idToken, audience string) error {
```
mpminardi, line 166: Note: it might also be worth looking into the more "native" options for [configuration validation in the framework](https://developer.hashicorp.com/terraform/plugin/framework/providers/validate-configuration) as a followup if this ends up being easier to reason about / add new validations.

mpminardi (approved):

PR #739, opened 2026-05-12
all: migrate from terraform-plugin-sdk to terraform-plugin-testing

alexwlchan:
The testing parts of the plugin SDK have been extracted into a new package which has the same API. We need to update the imports and a couple of tests where the new package is stricter.

Updates tailscale/corp#37221

### DIFF
--- a/tailscale/resource_acl_test.go
+++ b/tailscale/resource_acl_test.go
@@ -12,8 +12,10 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+
+	"github.com/tailscale/hujson"
 
 	"tailscale.com/client/tailscale/v2"
 )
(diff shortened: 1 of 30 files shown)


alexwlchan:
Successful acceptance test run: https://github.com/tailscale/corp/actions/runs/25740093677

tailscale/resource_acl_test.go
```diff
@@ -216,10 +218,41 @@ func TestAccACL(t *testing.T) {
...
 			},
 			{
-				ResourceName:            resourceName,
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"overwrite_existing_content"},
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: false,
+				ImportStateCheck: func(s []*terraform.InstanceState) error {
```
mpminardi, line 224: Was this change needed due to difference in whitespace in the ACL being more problematic when testing with the new package?

mpminardi (approved):

tailscale/resource_acl_test.go
alexwlchan: Yeah, the test fails without it. I'm guessing it's something to do with when the plan modifier does/doesn't run during the `ImportStateVerify` checks.

```
    resource_acl_test.go:186: Step 3/3 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.

          map[string]string{
          	"acl": strings.Join({
        - 		`			{`,
        - 		`				// Tag owners.`,
        - 		`				"TagOwners": {`,
        - 		`					"tag:example": [`,
        - 		`						"autogroup:member"`,
        - 		`					]`,
        - 		`				},`,
        - 		`			}`,
        + 		"{",
        + 		`	// Tag owners.`,
        + 		`	"TagOwners": {`,
        + 		`		"tag:example": [`,
        + 		`			"autogroup:member",`,
        + 		`		],`,
        + 		`	},`,
        + 		"}",
          		... // 1 identical line
          	}, "\n"),
          }
```

PR #747, opened 2026-05-12
resource_device_tags: fix drift when tags are empty

zofrex:
Without this patch, we will always get a diff when `tags` is an empty list, e.g.:

```terraform
resource "tailscale_device_tags" "test_tags" {
  device_id = data.tailscale_device.test_device.id
  tags = []
}
```

Produces this diff on apply every time:

```
  # tailscale_device_tags.test_tags will be updated in-place
  ~ resource "tailscale_device_tags" "test_tags" {
      ~ id        = "1333080874923104" -> (known after apply)
      + tags      = []
        # (1 unchanged attribute hidden)
    }
```

This patch sets a default value of `[]`, the same as how `resource_tailnet_key` works.

Updates tailscale/corp#37032

### DIFF
--- a/tailscale/resource_device_tags_test.go
+++ b/tailscale/resource_device_tags_test.go
@@ -6,6 +6,7 @@ package tailscale
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"os"
 	"reflect"
 	"testing"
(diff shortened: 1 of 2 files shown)


tailscale/resource_device_tags_test.go
```diff
@@ -19,12 +19,23 @@ import (
...
 
+	const testDeviceTagsCreateEmpty = `
+		data "tailscale_device" "test_device" {
+			name = "%s"
+			wait_for = "60s"
+		}
+
+		resource "tailscale_device_tags" "test_tags" {
+			device_id = data.tailscale_device.test_device.id
+			tags = []
```
zofrex, line 24: This is an annoying test because you need to create a new device, or re-auth the test device, every time you run the test — you cannot remove all tags from a node except via re-auth, and later in the test we add some tags. Don't think there's much we can do about it, though.

neinkeinkaffee: Is the idea to do this manually, e.g. by restarting devcontrol? I can think of at least one other instance (the logic deleting old IDs from state on 404) where we didn't write an acceptance test because the it would have involved setting up a way to repopulate the test tailnet with new devices.

Would a unit test actually be a valid alternative here, similar to the ones for the device data source ([here](https://github.com/tailscale/terraform-provider-tailscale/blob/4b5bb8d671e0a88312381a7dedae1c1e7c5d26eb/tailscale/data_source_device_test.go#L128) and following)? I imagine we could set up the test server to return tag=nil and then test that the plan comes out empty.

zofrex: One option @mpminardi and I were discussing is maybe there should be a way to create a device in the database without needing to spin up a client and log in. We can do it in unit tests, but not from functional tests like these (outside of the seed data step). It's something I've run into outside of terraform testing as well.

mpminardi: I think a unit test would be the move here at least in the short term to avoid friction with having to do more manual intervention to make the tests repeatable locally. 

It will be good to have a better story around this for being able to validate this (and other cases) using an acceptance test in the future (and for other cases outside of Terraform testing like @zofrex mentioned) so I think we should do some more investigation around creating a DB only device as a follow-up.

zofrex: 👍 changed this one to a unit test

mpminardi (approved):

neinkeinkaffee (approved):

PR #750, opened 2026-05-13
resource_logstream_configuration: fix drift when upload_period_minutes is 0

zofrex:
Without this patch, we will always get a diff when `upload_period_minutes` is `0`. E.G. this terraform:

```terraform
resource "tailscale_logstream_configuration" "test_logstream_configuration" {
		log_type         = "configuration"
		destination_type = "panther"
		url              = "https://example.com"
		token            = "some-token"
		upload_period_minutes = 0
}
```

Produces this diff on apply every time:

```
  # tailscale_logstream_configuration.test_logstream_configuration will be updated in-place
  ~ resource "tailscale_logstream_configuration" "test_logstream_configuration" {
        id                    = "configuration"
      + upload_period_minutes = 0
        # (6 unchanged attributes hidden)
    }
```

This patch sets a default value of 0 for this. Confirmed that 0 is equivalent to unset as far as the API is concerned.

Updates tailscale/corp#37032


alexwlchan (approved):
🕐 0️⃣

mpminardi (approved):

PR #755, opened 2026-05-14
resource_dns_configuration: error if no nameservers provided for splitdns

zofrex:
Without this patch, the following terraform plans successfully:

```terraform
resource "tailscale_dns_configuration" "test_configuration" {
  split_dns {
    domain             = "foo.example.com"
    # nameservers {
    #   address            = "1.1.1.3"
    # }
  }
}
```

Which then errors on applying:

```
╷
│ Error: Failed to set DNS configuration
│
│   with tailscale_dns_configuration.test_configuration,
│   on main.tf line 63, in resource "tailscale_dns_configuration" "test_configuration":
│   63: resource "tailscale_dns_configuration" "test_configuration" {
│
│ invalid character 'I' looking for beginning of value
╵
```

The schema has a validator:

```
	Validators: []validator.List{
		listvalidator.SizeAtLeast(1),
	},
```

But the docs for `SizeAtLeast` explains why this has no effect:

> Null (unconfigured) and unknown (known after apply) values are skipped.

The previous version of the plugin provides a nice error message on plan here:

```
│ Error: Insufficient nameservers blocks
│
│   on main.tf line 64, in resource "tailscale_dns_configuration" "test_configuration":
│   64:   split_dns {
│
│ At least 1 "nameservers" blocks are required.
```

We could use the `listvalidator.IsRequired` validator, which even suggests it's an appropriate validator for the scenario:

```
// This validator is equivalent to the `Required` field on attributes and is only
// practical for use with `schema.ListNestedBlock`
```

But that produces a much less helpful error message:

```
│ Error: Invalid Block
│
│   with tailscale_dns_configuration.test_configuration,
│   on main.tf line 64, in resource "tailscale_dns_configuration" "test_configuration":
│   64:   split_dns {
│
│ Block split_dns[0].nameservers must have a configuration value as the provider has marked it as required
```

Instead, I've added a custom validator, `AtLeastOneBlockRequired`, that checks that the provided `ListNestedBlock` is not null and has length at least 1, and provides this error message if it does not:

```
│ Error: Insufficient nameservers blocks
│
│   with tailscale_dns_configuration.test_configuration,
│   on main.tf line 64, in resource "tailscale_dns_configuration" "test_configuration":
│   64:   split_dns {
│
│ At least 1 "nameservers" blocks are required.
╵
```

Updates tailscale/corp#37032

### DIFF
--- a/tailscale/resource_dns_configuration.go
+++ b/tailscale/resource_dns_configuration.go
@@ -8,7 +8,6 @@ import (
 	"slices"
 	"strings"
 
-	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
@@ -107,8 +106,9 @@ func (r *dnsConfigurationResource) Schema(_ context.Context, _ resource.SchemaRe
 					Blocks: map[string]schema.Block{
 						"nameservers": schema.ListNestedBlock{
 							Description: "Set the nameservers used by devices on your network to resolve DNS queries.",
+
 							Validators: []validator.List{
-								listvalidator.SizeAtLeast(1),
+								AtLeastOneBlockRequired(),
 							},
 							NestedObject: schema.NestedBlockObject{
 								Attributes: map[string]schema.Attribute{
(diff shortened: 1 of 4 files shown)


tailscale/resource_dns_configuration.go
```diff
@@ -311,3 +306,40 @@ func (r *dnsConfigurationResource) updateDNSConfiguration(ctx context.Context, d
 		diags.AddError("Failed to set DNS configuration", err.Error())
 	}
 }
+
+var _ validator.List = atLeastOneBlockRequiredValidator{}
```
mpminardi: Can do this as a followup, but can we move this to `validators.go` and add a test if possible?

mpminardi (approved):

PR #760, opened 2026-05-14
resource_contacts: error if contacts are not set exactly once each

zofrex:
Similar to #755, but a `Set` not a `List`.

Without this patch, the following terraform plans successfully:

```terraform
resource "tailscale_contacts" "test_contacts" {
  account {
    email = "account@example.com"
  }

  support {
    email = "support@example.com"
  }

  # security {
  #   email = "security@example.com"
  # }
}
```

This is not permitted in the previous version of the provider:

```
│ Error: Insufficient security blocks
│
│   on main.tf line 31, in resource "tailscale_contacts" "test_contacts":
│   31: resource "tailscale_contacts" "test_contacts" {
│
│ At least 1 "security" blocks are required.
```

The schema has validators:

```
	Validators: []validator.Set{
		setvalidator.SizeAtLeast(1),
		setvalidator.SizeAtMost(1),
	},
```

But the docs for `SizeAtLeast` explains why this has no effect (same as in #755):

> Null (unconfigured) and unknown (known after apply) values are skipped.

We could use the `listvalidator.IsRequired` validator, which even suggests it's an appropriate validator for the scenario:

```
// This validator is equivalent to the `Required` field on attributes and is only
// practical for use with `schema.ListNestedBlock`
```

But that produces a slightly less helpful error message:

```
│ Error: Invalid Block
│
│   with tailscale_contacts.test_contacts,
│   on main.tf line 31, in resource "tailscale_contacts" "test_contacts":
│   31: resource "tailscale_contacts" "test_contacts" {
│
│ Block security must have a configuration value as the provider has marked it as required
```

Additionally, the error message for supplying too many blocks, rather than not supplying one, also got a little worse. The terraform:

```terraform
resource "tailscale_contacts" "test_contacts" {
  account {
    email = "account@example.com"
  }

  support {
    email = "support@example.com"
  }

  security {
    email = "security1@example.com"
  }

  security {
    email = "security2@example.com"
  }
}
```

On the previous provider results in:

```
│ Error: Too many security blocks
│
│   on main.tf line 44, in resource "tailscale_contacts" "test_contacts":
│   44:   security {
│
│ No more than 1 "security" blocks are allowed
```

But on the latest we get:

```
│ Error: Invalid Attribute Value
│
│   with tailscale_contacts.test_contacts,
│   on main.tf line 31, in resource "tailscale_contacts" "test_contacts":
│   31: resource "tailscale_contacts" "test_contacts" {
│
│ Attribute security set must contain at most 1 elements, got: 2
```

This PR adds a custom validator, `ExactlyOneBlockRequiredSet`, that checks that the provided `SetNestedBlock` is not null and has length exactly 1. If it is not set, or length 0, it produces this error message:

```
│ Error: No security block
│
│   with tailscale_contacts.test_contacts,
│   on main.tf line 31, in resource "tailscale_contacts" "test_contacts":
│   31: resource "tailscale_contacts" "test_contacts" {
│
│ A "security" block is required.
```

And if the length is more than 1, it produces _this_ error message:

```
│ Error: Too many security blocks
│
│   with tailscale_contacts.test_contacts,
│   on main.tf line 31, in resource "tailscale_contacts" "test_contacts":
│   31: resource "tailscale_contacts" "test_contacts" {
│
│ Only one "security" block is allowed.
```

There is a slight quirk where this is allowed:

```terraform
  security {
    email = "security@example.com"
  }

  security {
    email = "security@example.com"
  }
```

Even though this is not:

```
  security {
    email = "security@example.com"
  }

  security {
    email = "security2@example.com" # N.B: this one is different
  }
```

I think this is because it is a `Set` and not a `List` so equal blocks are de-duped?

Updates tailscale/corp#37032

### DIFF
--- a/tailscale/resource_contacts.go
+++ b/tailscale/resource_contacts.go
@@ -6,7 +6,6 @@ package tailscale
 import (
 	"context"
 
-	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
@@ -69,8 +68,7 @@ func (r *contactsResource) Schema(_ context.Context, _ resource.SchemaRequest, r
 					},
 				},
 				Validators: []validator.Set{
-					setvalidator.SizeAtLeast(1),
-					setvalidator.SizeAtMost(1),
+					ExactlyOneBlockRequiredSet(),
 				},
 			},
 			"support": schema.SetNestedBlock{
@@ -84,8 +82,7 @@ func (r *contactsResource) Schema(_ context.Context, _ resource.SchemaRequest, r
 					},
 				},
 				Validators: []validator.Set{
-					setvalidator.SizeAtLeast(1),
-					setvalidator.SizeAtMost(1),
+					ExactlyOneBlockRequiredSet(),
 				},
 			},
 			"security": schema.SetNestedBlock{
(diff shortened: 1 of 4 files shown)


mpminardi (approved):
Thank you for the thorough writeup!

PR #763, opened 2026-05-18
resource_acl: validate policy file at plan time

### LINKED ISSUE #762: tailscale_acl: validate policy file on plan

I have a CI workflow set up to do planning on PR, but it failed after merge because there was a test in my policy file that wasn't validated before merge.

On apply the only error I got was `400`, which also made it hard to figure out what went wrong

I'd like the policy file to get validated while planning.

bouk:
## Summary

- Adds `ModifyPlan` to `tailscale_acl` that calls the policy file Validate API at plan time.
- Updates the resource description accordingly.

Fixes #762.

### DIFF
--- a/docs/resources/acl.md
+++ b/docs/resources/acl.md
@@ -3,14 +3,14 @@ page_title: "tailscale_acl Resource - terraform-provider-tailscale"
 subcategory: ""
 description: |-
   The acl resource allows you to configure a Tailscale policy file. See https://tailscale.com/kb/1395/tailnet-policy-file for more information. Note that this resource will completely overwrite existing policy file contents for a given tailnet.
-  If tests are defined in the policy file (the top-level "tests" section), policy file validation will occur before creation and update operations are applied.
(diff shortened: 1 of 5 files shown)


alexwlchan:
Thanks for this @bouk! The acceptance tests aren't working as I'd expect – if I remove the mocked responses and replace the ACL with one that fails tests, I get the same unhelpful "Invalid ACL" message. That might be a problem in the Go client, which I'm going to check next.

Additionally, we should make sure whatever nice error handling we add here is also supported in the apply stage, for anybody who skips the plan.

alexwlchan:
I understand why this isn't working! There's a bug in the Go client SDK: https://github.com/tailscale/tailscale-client-go-v2/issues/100

We'll fix that, then we can revisit this patch.

neinkeinkaffee (approved):

neinkeinkaffee:
Thank you for the contribution @bouk! With @alexwlchan's  fix of the Go client SDK this is now good to be rebased and merged.

Issue #764, opened 2026-05-19
`tailscale_tailnet_key` key nulled on state refresh


horjulf:
**Describe the bug**

`v0.29` with https://github.com/tailscale/terraform-provider-tailscale/pull/726 introduced a change in state for `tailscale_tailnet_key`, after initial resource creation, on the next state read, `key` becomes `null`.

For example, it breaks use-cases where the key is injected into user-data for client auth.

```
# module.tailscale.tailscale_tailnet_key.default has changed
  ~ resource "tailscale_tailnet_key" "default" {
        id            = "...."
      - key           = (sensitive value) -> null
        tags          = [
            "tag:...",
        ]
        # (9 unchanged attributes hidden)
    }
```

**To Reproduce**
Steps to reproduce the behaviour:
1. Create `tailscale_tailnet_key`
2. Apply
3. Plan/apply again

**Expected behaviour**

`key` should be retained in state.

horjulf:
cc/ @zofrex, at first glance it seems the change in behavior is triggered by the plan modifier `UseStateForUnknown`.

michaelbeaumont:
It also breaks things on upgrade with an NPE since after a not found error:

https://github.com/tailscale/terraform-provider-tailscale/blob/f4499a002d78f92cd6f057f9218e7d87c6918bed/tailscale/resource_tailnet_key.go#L291

immediately the nil value is read:

https://github.com/tailscale/terraform-provider-tailscale/blob/f4499a002d78f92cd6f057f9218e7d87c6918bed/tailscale/resource_tailnet_key.go#L299

zofrex:
@michaelbeaumont I wasn't able to reproduce the NPE aspect of this. Can you share reproduction steps?

Edit: NVM, @mpminardi shared reproduction steps with me ("the key needs to have been actually deleted vs. revoked").

mpminardi:
@horjulf @michaelbeaumont thank you for flagging this! Currently working on validating a fix for these issues.

mpminardi:
Hey folks! We've released `v0.29.1` which should resolve these issues.

horjulf:
Thank you for the quick fix and release!

PR #765, opened 2026-05-19
resource_tailnet_key: fix not found handling and key state drift

mpminardi:
Do a resp.State.RemoveResource call and early return when the key is not
found to properly attempt to recreate the key instead of later running
into a nil pointer.

Remove logic that set the key property in the read and update methods to
prevent state drift of this attribute.

Updates https://github.com/tailscale/terraform-provider-tailscale/issues/764

### DIFF
--- a/tailscale/resource_tailnet_key.go
+++ b/tailscale/resource_tailnet_key.go
@@ -110,7 +110,7 @@ func (t *tailnetKeyResource) Schema(_ context.Context, _ resource.SchemaRequest,
 				Optional:      true,
 				Computed:      true,
 				Description:   "The expiry of the key in seconds. Defaults to `7776000` (90 days).",
-				PlanModifiers: []planmodifier.Int64{int64planmodifier.RequiresReplace()},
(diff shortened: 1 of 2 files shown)


zofrex (approved):
A test would be good of course but doesn't need to block merging this. Tested manually and it fixes the issue.

Of course, if someone already has an empty value for `key` in their state due to this bug, this won't recover it, they will need to recreate the key, but I don't think there's anything we can do about that.

zofrex:
Also fixes the NPE in my testing.

mpminardi:
> A test would be good of course but doesn't need to block merging this. Tested manually and it fixes the issue.
> 
> Of course, if someone already has an empty value for `key` in their state due to this bug, this won't recover it, they will need to recreate the key, but I don't think there's anything we can do about that.

Ended up adding some minimal tests with a reproducer for the state drift piece of things, successful acceptance test run up at https://github.com/tailscale/corp/actions/runs/26124449936

PR #768, opened 2026-05-22
resource_device_key: do nothing on delete

### LINKED ISSUE #401: Deleting tailscale_device_key resources results in key expiry being disabled
When the resource is deleted, the provider calls the API with an empty `tsclient.DeviceKey{}`, which results in `KeyExpiryDisabled` being set to false on the server.

It's not clear that this is the behavior we want when deleting device key resources.

zofrex:
Fixes #401 by changing deletion to a no-op. Also add a warning about that:

```
│ Warning: Resource Destruction Considerations
│
│ Destroying this resource will only remove the resource from the Terraform state and
│ will not undo or change the device's key settings. Use a tailscale_device_key resource
│ to explicitly change the key expiry setting, or manually update it via the admin
│ console.
```

### DIFF
--- a/tailscale/resource_device_key.go
+++ b/tailscale/resource_device_key.go
@@ -19,6 +19,7 @@ var (
 	_ resource.Resource                = &deviceKeyResource{}
 	_ resource.ResourceWithConfigure   = &deviceKeyResource{}
 	_ resource.ResourceWithImportState = &deviceKeyResource{}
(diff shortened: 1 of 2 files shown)


alexwlchan (approved):
🔑

zofrex:
(per Mario) big enough behaviour change that we want to hold off until v1 on this one.

Issue #770, opened 2026-05-26
tailscale_tailnet_key: recreate_if_invalid not respected when key is not found


mpminardi:
As part of https://github.com/tailscale/terraform-provider-tailscale/pull/765 logic was added to handle removing the resource from state when the controlplane returns a 404.

This change missed checking `recreate_if_invalid` before doing the above which is causing keys to be recreated when they should not be as per that setting.

PR #771, opened 2026-05-26
resource_tailnet_key: fix erroneous key recreation in not found case

mpminardi:
As part of https://github.com/tailscale/terraform-provider-tailscale/pull/765 logic was added to remove the resource from state if the remote key was not found.

This original change added a removal of the resource from state directly in the not found handling instead of letting this be handled by the existing ModifyPlan logic.

This change removes the call to resp.State.RemoveResource and instead leaves this to be handled by ModifyPlan.

Updates https://github.com/tailscale/terraform-provider-tailscale/issues/770

### DIFF
--- a/tailscale/resource_tailnet_key.go
+++ b/tailscale/resource_tailnet_key.go
@@ -239,7 +239,7 @@ func (t *tailnetKeyResource) Read(ctx context.Context, req resource.ReadRequest,
 	key, err := t.Client.Keys().Get(ctx, state.ID.ValueString())
 	if tailscale.IsNotFound(err) {
 		state.Invalid = types.BoolValue(true)
-		resp.State.RemoveResource(ctx)
+		resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
 		return
 	} else if err != nil {
 		resp.Diagnostics.AddError("Failed to fetch key", fmt.Sprintf("Error reading tailnet key with id %q: %s", state.ID, err.Error()))
--- a/tailscale/resource_tailnet_key_test.go
+++ b/tailscale/resource_tailnet_key_test.go
@@ -103,9 +103,32 @@ func setKeyStep(reusable bool, recreateIfInvalid string) resource.TestStep {
 	}
 }
 
-func checkInvalidKeyRecreated(description string, reusable bool, recreateIfInvalid string, wantRecreated bool) resource.TestStep {
+type checkInvalidKeyRecreatedParams struct {
+	description       string
+	reusable          bool
+	recreateIfInvalid string
+	keyNotFound       bool
+	wantRecreated     bool
+}
+
+func checkInvalidKeyRecreated(params checkInvalidKeyRecreatedParams) resource.TestStep {
+	description := params.description
+	reusable := params.reusable
+	recreateIfInvalid := params.recreateIfInvalid
+	keyNotFound := params.keyNotFound
+	wantRecreated := params.wantRecreated
+
 	return resource.TestStep{
 		PreConfig: func() {
+			if keyNotFound {
+				newKey := testTailnetKeyStructWithID("new-key-id", reusable)
+				testServer.SetResponses([]TestResponse{
+					{Code: http.StatusNotFound},
+					{Code: http.StatusOK, Body: newKey},
+				})
+				return
+			}
+
 			oldKey := testTailnetKeyStructWithID("old-key-id", reusable)
 			oldKey.Invalid = true
 			newKey := testTailnetKeyStructWithID("new-key-id", reusable)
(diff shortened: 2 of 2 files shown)


tailscale/resource_tailnet_key_test.go
```diff
@@ -103,9 +103,30 @@ func setKeyStep(reusable bool, recreateIfInvalid string) resource.TestStep {
 	}
 }
 
-func checkInvalidKeyRecreated(description string, reusable bool, recreateIfInvalid string, wantRecreated bool) resource.TestStep {
+type checkInvalidKeyRecreatedParams struct {
```
mpminardi, line 106: Extracted the arguments into a params struct since it was getting hard to track down what was going on in calls to `checkInvalidKeyRecreated`

tailscale/resource_tailnet_key_test.go
```diff
@@ -147,40 +168,69 @@ func checkInvalidKeyRecreated(description string, reusable bool, recreateIfInval
...
 			// Now make it a single-use key.
 			setKeyStep(false, ""),
-			checkInvalidKeyRecreated("Confirm that the single-use key is not recreated", false, "", false),
+			checkInvalidKeyRecreated(checkInvalidKeyRecreatedParams{
+				description:       "Confirm that the single-use key is not recreated",
+				reusable:          false,
+				recreateIfInvalid: "",
+				wantRecreated:     false,
+			}),
+			checkInvalidKeyRecreated(checkInvalidKeyRecreatedParams{
```
mpminardi, line 194: This is the new test step that validates we aren't recreating when the key is not found. Verified that this fails before the logic change.

tailscale/resource_tailnet_key.go
```diff
@@ -238,8 +238,10 @@ func (t *tailnetKeyResource) Read(ctx context.Context, req resource.ReadRequest,
 
 	key, err := t.Client.Keys().Get(ctx, state.ID.ValueString())
 	if tailscale.IsNotFound(err) {
-		state.Invalid = types.BoolValue(true)
-		resp.State.RemoveResource(ctx)
+		if shouldRecreateIfInvalid(state.Reusable.ValueBool(), state.RecreateIfInvalid.ValueString()) {
+			state.Invalid = types.BoolValue(true)
+			resp.State.RemoveResource(ctx)
+		}
```
zofrex: Two questions:
1. Should setting invalid to true be conditional on `shouldRecreateIfInvalid`? I don't think it needs to be, as we handle that solely in `ModifyPlan` now, we don't need to fake it on read (and further down, we do set it to true unconditionally if that's what we get back from the API)
2. Do we need to actually remove the resource from the state? The early return is what fixed the prior bug, if we set invalid to true and return I _think_ everything does the right thing?

```suggestion
		state.Invalid = types.BoolValue(true)
		resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
```

(ed: it does seem to work, with the addition of setting the state before returning)

tailscale/resource_tailnet_key.go
```diff
@@ -238,8 +238,10 @@ func (t *tailnetKeyResource) Read(ctx context.Context, req resource.ReadRequest,
 
 	key, err := t.Client.Keys().Get(ctx, state.ID.ValueString())
 	if tailscale.IsNotFound(err) {
-		state.Invalid = types.BoolValue(true)
-		resp.State.RemoveResource(ctx)
+		if shouldRecreateIfInvalid(state.Reusable.ValueBool(), state.RecreateIfInvalid.ValueString()) {
+			state.Invalid = types.BoolValue(true)
```
zofrex: This won't have any effect as it's not being written back

tailscale/resource_tailnet_key.go
mpminardi: Good catch! Yeah I think that just setting this to invalid + letting the `ModifyPlan` handle the actual the recreate bits should work. Will adjust this shortly.

tailscale/resource_tailnet_key.go
```diff
@@ -239,7 +239,6 @@ func (t *tailnetKeyResource) Read(ctx context.Context, req resource.ReadRequest,
 	key, err := t.Client.Keys().Get(ctx, state.ID.ValueString())
 	if tailscale.IsNotFound(err) {
 		state.Invalid = types.BoolValue(true)
```
zofrex, line 241: Still having no effect due to returning without saving

mpminardi: Whoops! Will adjust

zofrex (changes requested):
In my manual testing this doesn't handle the 404 case, the resource in my state file still has `"invalid": false` and it isn't being recreated. I think that means we need another unit test here!

mpminardi:
> In my manual testing this doesn't handle the 404 case, the resource in my state file still has `"invalid": false` and it isn't being recreated. I think that means we need another unit test here!

Oh interesting, will poke around some more on my end locally.

tailscale/resource_tailnet_key.go
mpminardi: Have a call to `resp.State.Set` here now

tailscale/resource_tailnet_key_test.go
```diff
@@ -147,40 +170,79 @@ func checkInvalidKeyRecreated(description string, reusable bool, recreateIfInval
...
 			// A single-use key with recreate=always, should be recreated.
 			setKeyStep(false, "always"),
-			checkInvalidKeyRecreated("A single-use key with recreate=always, should be recreated", false, "always", true),
+			checkInvalidKeyRecreated(checkInvalidKeyRecreatedParams{
+				description:       "A single-use key with recreate=always, should be recreated",
+				reusable:          false,
+				recreateIfInvalid: "always",
+				wantRecreated:     true,
+			}),
+			// A single-use key with recreate=always, should be recreated when the key is not found.
```
mpminardi, line 210: @zofrex added this to test that a 404 properly recreates when `recreateIfInvalid` is actually set.

zofrex (approved):

PR #778, opened 2026-06-08
dns_configuration: extract a reconcileSplitDNS function

alexwlchan:
This is a leftover TODO from the migration to the plugin SDK.

Updates #cleanup


neinkeinkaffee (commented):
Might it be worthwhile, now that we can, to add a small unit test documenting what the reconciliation does (even if it's for now just one test case that more or less a copies paste of the update scenario from the resource's acc test)?

alexwlchan:
Test cases added!

neinkeinkaffee (approved):

PR #791, opened 2026-07-13
tailscale_federated_identity, tailscale_posture_integration: import empty string fields cleanly

### LINKED ISSUE #790: Importing federated_identity or posture_integration with empty string fields plans a spurious update
### Summary

Importing an existing `tailscale_federated_identity` that has no description shows a spurious in-place update on the next plan, whose only change is `updated_at -> (known after apply)`. `tailscale_posture_integration` has the same problem on `cloud_id`, `client_id`, and `tenant_id`.

### Cause

These are `Optional+Computed` attributes with a schema `Default` of `""`, but their Read path runs the API value through `CoalesceStringEmptyOrNull`, which returns null when the value is empty. On import the prior state is null, so an empty value reads back as null while the schema default plans `""`. Terraform renders the null-vs-"" mismatch as an update, and the computed `updated_at` then goes `(known after apply)`.

### Reproduce

1. Have a `tailscale_federated_identity` with no `description`.
2. `terraform import tailscale_federated_identity.example <id>`
3. `terraform plan`: the resource updates in-place, with `updated_at -> (known after apply)` as the only diff.

### Interim workaround

Set a non-empty `description` on the resource and apply it before importing. Once the field is non-empty on both the config and the API, the import is clean.

### Affected

`tailscale_federated_identity` (`description`) and `tailscale_posture_integration` (`cloud_id`, `client_id`, `tenant_id`). `tailscale_oauth_client`'s `description` is on the SDKv2 path and is not affected.

keeleysam:

Importing an existing `tailscale_federated_identity` that has no description shows a spurious in-place update on the first plan. `description` is `Optional+Computed` with a schema `Default` of `""`, but its Read path runs the API value through `CoalesceStringEmptyOrNull`, which returns a null `StringValue` when the value is empty. On import the prior model value is null, so an empty description reads back as null while the default plans `""`. That leaves a permanent null-vs-"" mismatch, which Terraform renders as an update. Because it plans an update, the computed `updated_at` also goes `(known after apply)`, so importing a batch of identities that never had descriptions looks like they are all changing when nothing actually is.

`tailscale_posture_integration` has the same shape and the same bug on `cloud_id`, `client_id`, and `tenant_id` (all `Optional+Computed` with `Default` `""`, all read through `CoalesceStringEmptyOrNull`).

The fix sets these with `types.StringValue` so an empty value stays `""`, matching the schema default, and imports cleanly. `StringValueNullIfEmpty` is unchanged and still used for the fields that are genuinely null-when-empty and have no default (logstream `s3_authentication_type`, webhook `provider_type`).

Fixes #790

- Added a unit test that imports a `tailscale_federated_identity` with no description and asserts the imported state matches (it fails without the fix on `ImportStateVerify`, passes with it). The posture case is already covered by `TestAccTailscalePostureIntegration`, which imports a Falcon integration with no `tenant_id` and runs `ImportStateVerify`.
- `tailscale_oauth_client`'s `description` is on the SDKv2 code path, which does not exhibit this, so it is untouched.
- After this change `CoalesceStringEmptyOrNull` has no remaining callers. I left it in place since it is the correct helper for its documented pattern (a null-when-empty field with no default); its two `Default: ""` call sites were the wrong use. Happy to drop it if you would rather not carry an unused helper.
- Interim workaround for anyone hitting this before a release: set a non-empty `description` (and apply) before importing, so the field is non-empty on both sides and imports clean.
- `go test ./tailscale/`, `go build`, `go vet`, and `gofmt` are all clean.


tailscale/resource_federated_identity.go
```diff
@@ -267,7 +267,11 @@ func (r *federatedIdentityResource) populateFromKey(ctx context.Context, data *f
 	var diags diag.Diagnostics
 
 	data.ID = types.StringValue(key.ID)
-	data.Description = CoalesceStringEmptyOrNull(data.Description, key.Description)
+	// description has a schema Default of "", so represent an empty description as
+	// "" rather than null. Otherwise an empty description reads back as null on
+	// import, leaving a permanent null-vs-"" mismatch that plans as a spurious
+	// in-place update (which also flips the computed updated_at to known-after-apply).
+	data.Description = types.StringValue(key.Description)
```
mpminardi, line 274: I think we can drop the leading comment here and leave the comment in the regression test since this won't be a special case after this fix!

tailscale/resource_posture_integration.go
```diff
@@ -116,9 +116,13 @@ func (p *postureIntegrationResource) Read(ctx context.Context, req resource.Read
...
 	state.ID = types.StringValue(integration.ID)
 	state.PostureProvider = types.StringValue(string(integration.Provider))
-	state.CloudID = CoalesceStringEmptyOrNull(state.CloudID, integration.CloudID)
-	state.ClientID = CoalesceStringEmptyOrNull(state.ClientID, integration.ClientID)
-	state.TenantID = CoalesceStringEmptyOrNull(state.TenantID, integration.TenantID)
+	// cloud_id / client_id / tenant_id all have a schema Default of "", so represent
+	// empty values as "" rather than null. Otherwise an empty value reads back as
+	// null on import, leaving a permanent null-vs-"" mismatch (a spurious in-place
+	// update).
+	state.CloudID = types.StringValue(integration.CloudID)
```
mpminardi, line 123: Same RE: the leading comment here!

mpminardi:
Thank you for the PR @keeleysam ! And apologies for this having slipped through.

> After this change CoalesceStringEmptyOrNull has no remaining callers. I left it in place since it is the correct helper for its documented pattern (a null-when-empty field with no default); its two Default: "" call sites were the wrong use. Happy to drop it if you would rather not carry an unused helper.

I think we can drop the helper at this point! We've standardized instead on providing a default value and the helper is now vestigial.

PR #799, opened 2026-08-03
tailscale/resource_acl: preserve policy formatting across refresh

### LINKED ISSUE #798: tailscale_acl v0.29.x: semantic changes produce unreadable full-file formatting diff
Hi 👋 !

### Description

Since upgrading the Tailscale provider from 0.28.0 to 0.29.2, any semantic change to a `tailscale_acl` resource causes Terraform/OpenTofu to display almost the entire policy file as changed.

Most displayed changes are formatting differences between:

- the canonical HuJSON returned by the Tailscale API, using tabs and canonical line wrapping; and
- the original HuJSON supplied through `templatefile`, using spaces.

Whitespace-only changes are correctly suppressed. However, once there is a genuine policy change, the plan compares the API-formatted state against the raw configured string, making the plan practically impossible to review.

This worked correctly with provider 0.28.0.

### Provider and OpenTofu versions

- tailscale/tailscale: 0.29.2
- previous working provider: 0.28.0
- OpenTofu: 1.11.2
- Platform: darwin_arm64

### Configuration

We are using it similar to the [documented example](https://registry.terraform.io/providers/tailscale/tailscale/latest/docs/resources/acl#example-usage) but use the result of templatefile.

```hcl
resource "tailscale_acl" "main" {
  acl = templatefile("${path.module}/tailscale_acl.hujson", {
    installation_names = local.installation_names
  })
}
```

Simplified policy:

```hujson
{
  "hosts": {
    "vector": "100.64.0.1",
  },

  "ipsets": {
    "ipset:monitoring-endpoints": ["add host:vector"],
  },
}
```

Making this small semantic change:

```diff
 "hosts": {
   "vector": "100.64.0.1",
+  "vector-stage": "100.64.0.2",
 },

-"ipset:monitoring-endpoints": ["add host:vector"],
+"ipset:monitoring-endpoints": ["add host:vector", "add host:vector-stage"],
```

causes the plan to show the complete policy as removed and re-added, primarily changing tabs, spaces, wrapping, alignment, and trailing commas. Excerpt from the plan output:

```diff
-     "ipsets": {
-             "ipset:monitoring-endpoints": ["add host:vector"],
-     },
+   "ipsets": {
+     "ipset:monitoring-endpoints": ["add host:vector", "add host:vector-stage"],
+   },
```

### Expected behavior

The plan should show only the semantic HuJSON changes, as it did with provider 0.28.0.

### Regression analysis

Not done personally, the LLMs knew the version it worked with before and came up with this:

This appears to be a regression from the plugin-framework migration in: https://github.com/tailscale/terraform-provider-tailscale/pull/721

Provider 0.28.0 used `StateFunc` with `hujson.Format` and `DiffSuppressOnRefresh`: https://github.com/tailscale/terraform-provider-tailscale/blob/v0.28.0/tailscale/resource_acl.go#L64-L89

In 0.29.2, `Read` stores the raw API HuJSON in state: https://github.com/tailscale/terraform-provider-tailscale/blob/v0.29.2/tailscale/resource_acl.go#L86-L100

The new plan modifier suppresses the diff when canonical state and configuration are identical. If they differ semantically, however, it leaves the raw configured value in the plan: https://github.com/tailscale/terraform-provider-tailscale/blob/v0.29.2/tailscale/plan_modifiers.go#L64-L82

This means that any real change causes Terraform to compare the server-formatted state with the differently formatted configuration.

sidsri14:
## Summary

- model `acl` with a HuJSON custom value that treats canonical formatting differences as semantically equal
- let the Plugin Framework preserve the prior user-formatted value when `Read` receives equivalent canonical HuJSON from the API
- keep genuine policy changes byte-for-byte consistent with configuration instead of rewriting the configured value during planning
- cover formatting-only equality, real semantic changes, and the plan consistency regression

## Tests

- `go test -race ./tailscale -run TestAclHuJSON -count=1`
- `go test . -count=1`
- `go vet ./tailscale`
- `go build ./...`
- `go mod verify`

A full `go test ./tailscale` run reached the unrelated `TestDeviceRetry_EventualSuccess` Plugin Testing case, which was still downloading/unpacking Terraform through `hc-install` when the explicit 90-second timeout fired. The ACL-focused tests above pass under the race detector. `TF_ACC` tests were not run because they require a live Tailscale test account.

Fixes #798


neinkeinkaffee:
Hi @sidsri14,

Making semantic changes in the diff easier to spot is a good idea, and thanks for adding a test to illustrate the intent! 

Unfortunately, when I build the provider locally and plan an ACL change that isn’t canonical HuJSON (e.g. additional levels of indentation), I get an `Error: Provider produced invalid plan`. It seems that in its current form, the change doesn’t play well with Terraform’s [data consistency rules](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#caveats) that require consistency between config and plan.

sidsri14:
Thanks for reproducing this. You are right: canonicalizing a configured required value in the plan violates Terraform's data consistency rule. I replaced that approach in aaa0c20 with a HuJSON custom value implementing Plugin Framework semantic equality. Equivalent canonical API output now retains the prior user-formatted state during refresh, while a genuine change leaves the planned value exactly as configured. I also added a regression asserting that a non-canonical semantic change is not rewritten. The focused ACL tests pass under the race detector, and root tests, vet, build, and module verification pass.

PR #806, opened 2026-08-25
tailscale: remove the testify dependency

alexwlchan:
I keep getting Dependabot notifications to bump this dependency, but our test style is to use direct assertions instead. Rather than keep nagging, rewrite our handful of uses to our preferred style and remove the direct dep.

Updates #cleanup


alexwlchan:
Terraform acceptance tests: https://github.com/tailscale/corp/actions/runs/32845599476/job/97794432137

tailscale/data_source_device_test.go
```diff
@@ -51,26 +51,33 @@ func TestDeviceToMap(t *testing.T) {
...
-	assert.Equal(t, dev.MachineKey, m["machine_key"].(string))
-	assert.Equal(t, dev.NodeKey, m["node_key"].(string))
-	assert.Equal(t, dev.OS, m["os"].(string))
-	assert.Equal(t, dev.UpdateAvailable, m["update_available"].(bool))
-	assert.Equal(t, dev.TailnetLockError, m["tailnet_lock_error"].(string))
-	assert.Equal(t, dev.TailnetLockKey, m["tailnet_lock_key"].(string))
+	assertEqual := func(t *testing.T, want, got any) {
+		t.Helper()
+		if diff := cmp.Diff(want, got); diff != "" {
+			t.Errorf("wrong BuildRequest (-want +got):\n%s", diff)
```
bradfitz, line 57: to me, "assert" means t.Fatal, not t.Error

Why don't you switch to https://github.com/frankban/quicktest instead? That's the "blessed" consistent one we use elsewhere in the repos. Then we'd use the same verbs (check vs assert) to mean the same things.

Or rename this closure to "checkEqual" ?

tailscale/data_source_device_test.go
```diff
@@ -51,26 +51,33 @@ func TestDeviceToMap(t *testing.T) {
...
+		}
+	}
+
+	assertEqual(t, dev.Name, m["name"].(string))
```
bradfitz, line 61: this .(string) means you panic here instead of fail nicely with a nice error message if m["name"] is an unexpected type

PR #810, opened 2026-09-04
resource_tailnet_key: fix Update panic when key is not found

### LINKED ISSUE #809: tailscale_tailnet_key: Update panics with nil pointer dereference when key is not found
## Description

`tailnetKeyResource.Update` crashes with a nil pointer dereference when the underlying key returns a 404 from the Tailscale API (i.e., it's been fully purged, not just marked invalid) and the resource has an actual attribute change to apply (e.g. adding `recreate_if_invalid` to an existing resource, or any other config change).

`Read` correctly returns early after handling `tailscale.IsNotFound(err)`:

```go
func (t *tailnetKeyResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
	...
	key, err := t.Client.Keys().Get(ctx, state.ID.ValueString())
	if tailscale.IsNotFound(err) {
		state.Invalid = types.BoolValue(true)
		resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
		return
	}
	...
```

`Update` does not — it falls through and dereferences the nil `key`:

https://github.com/tailscale/terraform-provider-tailscale/blob/71abd07/tailscale/resource_tailnet_key.go#L290-L300

```go
	key, err := t.Client.Keys().Get(ctx, state.ID.ValueString())
	if tailscale.IsNotFound(err) {
		state.Invalid = types.BoolValue(true)
		// no return here
	} else if err != nil {
		...
		return
	} else {
		state.Invalid = types.BoolValue(false)
	}

	if key.KeyType != "auth" {  // key is nil when IsNotFound was true
```

## Reproduction

We hit this in practice on a `reusable = true` key with no `expiry` set (defaulting to the 90-day API default), which had genuinely expired and been purged from the Tailscale API by the time we tried adding `recreate_if_invalid = "never"` to an existing resource — a real attribute change, forcing Terraform to call `Update` rather than just refresh via `Read`.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x10 pc=...]

goroutine ... [running]:
github.com/tailscale/terraform-provider-tailscale/tailscale.(*tailnetKeyResource).Update(...)
	.../tailscale/resource_tailnet_key.go:300 +0x1a4
```

Version: `v0.29.2` (latest at time of writing). The bug is also present on `main`.

## Fix

I have a fix + a regression test ready (mirrors `Read`'s early return) and will open a PR shortly.

nicodono:
Fixes #809.

`Update` did not return after handling the `IsNotFound` case, unlike `Read`, so it fell through and dereferenced the nil `*tailscale.Key` it got back from a 404 — crashing with a nil pointer dereference.

This only reproduces when an actual attribute changes (forcing Terraform to call `Update` rather than just refresh via `Read`) while the key is no longer found — e.g. adding `recreate_if_invalid` to an existing resource whose key has since expired and been purged from the API. The existing test suite doesn't exercise `Update` at all in the not-found case (only `Read`/`ModifyPlan`), which is presumably how this slipped through.

## Changes

- `resource_tailnet_key.go`: `Update` now returns right after setting `Invalid` in the not-found branch, persisting the planned config (so any other attribute changes in the same apply, like `recreate_if_invalid`, still land in state) instead of falling through to dereference the nil key — mirrors `Read`'s existing handling of the same case.
- `resource_tailnet_key_test.go`: added `TestProvider_TailscaleTailnetKeyUpdateWhenNotFound`, a regression test that creates a key, then applies a real attribute change (`recreate_if_invalid` unset → `"never"`) while the mock server returns 404. Confirmed it reproduces the exact panic (same line, same trace) without the fix, and passes with it.

## Testing

- `go build ./...`: clean
- `go vet ./tailscale/...`: clean
- `gofmt -l`: clean
- `go test ./tailscale/...`: full package suite passes, including the new regression test


alexwlchan:
Thanks for the detailed bug report and the patch! 

The code looks good, but you need to sign the commit before we can approve/merge this: https://github.com/tailscale/terraform-provider-tailscale/pull/810/checks?check_run_id=101191666203

nicodono:
Fixed — amended the commit with a DCO sign-off and force-pushed. Check is green now, thanks for the quick look!

alexwlchan (approved):
Thank you!

PR #817, opened 2026-09-22
dns: expose exit node and local override options

### LINKED ISSUE #598: FR: add support for "use with exit node" to DNS resolvers
Currently you can toggle "Use with exit node" in the admin console for DNS resolvers (both Global and Restricted), which overrides the default behaviour to route all DNS queries to an exit node. 

Add a `use_with_exit_node` boolean to the `dns_split_nameservers` and `dns_nameservers` resources.

### LINKED ISSUE #789: `override_local_dns` missing in `tailscale_dns_preferences` resource
`override_local_dns` is present in the `tailscale_dns_configuration` resource but not in `tailscale_dns_preferences` resource (which only contains the `magic_dns` field)

Allow setting `override_local_dns` in `tailscale_dns_preferences` similar to how it can be set in `tailscale_dns_configuration`

**Additional context**
N/A

maxiscoding28:
## Summary

- add `use_with_exit_node` to `tailscale_dns_nameservers`
- add `use_with_exit_node` to `tailscale_dns_split_nameservers`
- add `override_local_dns` to `tailscale_dns_preferences`
- update tests, examples, and generated documentation

The global nameserver value applies to every global resolver. Each split-DNS resource applies its value to every resolver for that domain.

Fixes #598.
Fixes #789.

## Dependencies

This PR depends on these draft changes:

1. Control-plane API: https://github.com/tailscale/corp/pull/48709
2. Go API client: https://github.com/tailscale/tailscale-client-go-v2/pull/109

The control-plane change must deploy before this provider version is used. The client change must merge and receive a release before this PR is ready. This branch currently uses the client draft commit through a pseudo-version. Replace it with the released client version before merge.

The split-DNS resource uses the new partial PATCH method. It updates only its domain and avoids complete configuration replacement.

The global nameserver and DNS preference resources still use complete configuration updates. Do not change those resources concurrently with another DNS writer.

## Validation

- `go test -race ./...`
- live validation confirmed all three fields round-trip correctly
- live validation restored the complete DNS configuration to its original hash

The live validation covered the earlier complete-configuration implementation. The new split-DNS PATCH path requires the control-plane draft to deploy before live validation.


maxiscoding28:
Live validation completed with the provider built from commit `601d401`.

Validated:
- `tailscale_dns_split_nameservers.use_with_exit_node` created a temporary split-DNS entry with `useWithExitNode: true`.
- `tailscale_dns_nameservers.use_with_exit_node` changed both global resolvers to `true` and then restored them to `false`.
- `tailscale_dns_preferences.override_local_dns` changed from `true` to `false` and then restored to `true`.
- Unrelated DNS configuration remained unchanged during each operation.
- Terraform reported no changes after each update.
- The final complete DNS configuration hash matches the original hash.

maxiscoding28:
Follow-up: commit `e548b01` changes `tailscale_dns_split_nameservers` to use the new partial PATCH client from tailscale-client-go-v2#109. The earlier live validation covered the complete-configuration implementation. Live validation of the new partial PATCH path must wait for tailscale/corp#48709 to deploy.

maxiscoding28:
Validated end to end against a local dev control server running tailscale/corp#48709, with this branch's provider built locally and the client from tailscale-client-go-v2#109:

- All 7 DNS acceptance tests pass (`TestAccTailscaleDNS*`), including the SDK-to-framework migration tests against provider v0.29.2, which confirm upgrading doesn't produce a plan diff or flip settings.
- Manual `terraform apply` with the compiled provider binary: explicit `use_with_exit_node`/`override_local_dns` values round-trip; removing the attributes from config leaves existing flags untouched on the next apply (new resolvers default to false); plans stay clean; destroy removes everything.

Direct API checks against the same control server confirmed mixed string/object requests, option preservation, and 400s for invalid resolver objects.

